PR #5, opened 2020-11-26
Add the possibility to set return type.

oguimbal:
Consider the following use case:

```typescript
    const v = 'dt' as 'dt' | 'num' | 'nil';
    match(v)
      .with('dt', () => new Date())
      .with('num', () => 42)
      .with('nil', () => undefined)
      .run();
```

Unefortunately, the first type is picked, trigerring an error on `42` and `undefined`.

I know I could do something like this:
```typescript
match<typeof v, Date | number | undefined>(v)
```
... but I'm finding it a bit verbose, and forces me to store the matched value in a local variable (or have a named type for its value).

One option could be to make `PickReturnValue` use unions instead of just propagating the same type. However, I feel this could lead to unexpected quite complex types being infered when there actually is a mistake in one branch.

That's why this PR suggests another approach: Adding an `as<v>()` method that forces all branches to a given return type.

Tell me what you think of it ! (no problem if you dont want to merge it :) )

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -87,6 +87,9 @@ const builder = <a, b>(
       },
     ]).run(),
 
+  as: <c>(): Match<a, PickReturnValue<b, c>>  =>
+    builder<a, PickReturnValue<b, c>>(value, patterns),
+
   run: (): b => {
     const entry = patterns.find(({ test }) => test(value));
     if (!entry) {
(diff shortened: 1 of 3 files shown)


gvergnaud:
Hey!

Imo adding a method only to pass a type level argument doesn't feel very TS idiomatic, and I kind of prefer passing types argument this way `match<typeof v, Date | number | undefined>(v)`. If you think this is too verbose you can also give your first `with` function an explicit return type:

```ts
const v = 'dt' as 'dt' | 'num' | 'nil';
    match(v)
      .with('dt', (): Date | number | undefined => new Date())
      .with('num', () => 42)
      .with('nil', () => undefined)
      .run();
```

I'm going to close this PR, thanks for your suggestion, have a good weekend!

oguimbal:
No problem, that's a matter of taste, I guess :)

Issue #4, opened 2020-11-25
Exhaustiveness checking?


AndrewMorsillo:
Very cool library, thanks for creating it! Would it be possible to force exhaustiveness checking for union types? For example:

```
function getSomeString(unionType:  "foo" | "bar"): string{
     return match(unionType).with("foo", () => `A string result, but we won't check for "bar"!`).run()
}
```

This will not be a compile time error. If you write it as 

```
function getSomeString(unionType:  "foo" | "bar"): string{
    if(unionType === "foo"){
        return `A string result, but we won't check for "bar"!`
    } 
}
```

You get `TS2366: Function lacks ending return statement and return type does not include 'undefined'` because TS can verify that you haven't exhausted all of the cases.

It would be really nice if this library could take advantage of that. Failure to check all cases is a really easy bug to introduce!

gvergnaud:
👋 Hey, Thanks for your feedback!

I had a go at implementing exhaustive pattern matching when I first wrote this library ([Here is the branch if you are curious](https://github.com/gvergnaud/ts-pattern/tree/exhaustive-patterns)). I agree having exhaustive patterns would be immensely useful, but it's actually pretty hard to get right.

Support `when` predicate doesn't make our lives easier to implement it, because if the function given isn't a [Type Guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types) there is no way to narrow the return type to something more precise. I think that's why exhaustive pattern matching isn't turned on by default on languages like Haskell.

The second issue we face is with literal patterns. Consider this case:

```ts
const input: number
match(input)
    .with(2, () => 'two')
    .with(3, () => 'three')
    .run()
```

This block will not compile, because `2` 's inferred type turns out to be `number`, so the second `.with` will have `Exclude<number, number>` which evaluates to `never` as narrowed input type and which isn't compatible with a `3` pattern. Of course you could do `.with(2 as const, ...)` to mitigate this issue, but the fact that the default behavior is incorrect and confusing discouraged me to actually push this experiment to the main branch. I looked for ways to force Typescript to consider the first argument of `.with` as `const` or `readonly` by default, but I didn't find a way make it work. 

Another approach we could have would be to make exhaustive patterns an opt-in behavior by extending the API surface of the library, as @oguimbal suggested in this [PR suggested](https://github.com/gvergnaud/ts-pattern/pull/6) but I'd like to make sure the proposed API makes sense before merging it.

[...]

PR #6, opened 2020-11-26
Exhaustive pattern

oguimbal:
Implements exhaustive pattern matching (at least on types that extend string)

image

[edit] It turns out that i'm not the only one wishing that :)  cf. #4 , and i've seen that you've mentioned something like this in roadmap.md ... is that the kind of thing that you have in mind ?

nb: I introduced this new `otherwise()` function (which does essentially nothing except type checking) because the versatility of various pattern matching overloads is such that I think it is be impossible to implement exhaustive check on `.run()`,  (i'm looking right at you, predicates...), so I only see this exclusive matching as an opt-in behaviour.

(again, no pressure to merge this, I wont take it personally if you're going another way)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,7 +10,7 @@ import {
   not,
   select,
 } from './types/Pattern';
-import { Unset, Match, PickReturnValue } from './types/Match';
+import { Unset, Match, PickReturnValue, ExcludePattern } from './types/Match';
 
 /**
  * # Pattern matching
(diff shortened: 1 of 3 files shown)


gvergnaud:
Hey 👋 Thanks for contributing!

I posted a [comment in the related issue](https://github.com/gvergnaud/ts-pattern/issues/4#issuecomment-734950737), feel free to answer there so we keep the discussion centralised :)

gvergnaud:
Closing in favor of #7

PR #7, opened 2020-11-28
Add opt-in exhaustive pattern matching

gvergnaud:
Draft PR implementing exhaustive pattern matching as an opt-in feature.

## Example

```ts
type Option<a> = { kind: 'none' } | { kind: 'some'; value: a };

// This compiles:
match<Option<number>>({ kind: 'some', value: 3 })
        .exhaustive()
        .with({ kind: 'some' }, ({ value }) => value)
        .with({ kind: 'none' }, () => 0)
        .run();

// This doesn't
match<Option<number>>({ kind: 'some', value: 3 })
        .exhaustive()
        .with({ kind: 'some' }, ({ value }) => value)
        .run();
```

Exhaustive pattern matching will have several limitations, I'd like to write some good documentation about them to make sure people know what they can hope for when opting into this feature


tests/exhaustive-match.test.ts
```diff
@@ -0,0 +1,289 @@
...
+
+      match(input)
+        .exhaustive()
+        .with(['two', { kind: 'some' }], ([_, { value }]) => value.length)
+        .with(['two', { kind: 'none' }], () => 4)
+        .with([1, __], () => 3)
+        .with([3, __], () => 3)
+        // FIXME:  This case should work, because all cases of the Option are handled
+        // @ts-expect-error
+        .run();
```
gvergnaud, line 256: Pretty annoying gotcha, `Exclude`ing doesn't work on nested structures

oguimbal:
Haha nice, you've definitely superseeded my basic attempt (which was quite ok for my needs, as long as i'm concerned)

I like the idea of having removed predicates.

I'd add another test for this case (which fails as it)

```typescript
      type Input = { value: 'a' | 'b' }
      const input = { value: 'a' } as Input;

      match(input)
        .exhaustive()
        .with({ value: 'a' }, (x) => 1)
        // @ts-expect-error
        .run();

      match(input)
        .exhaustive()
        .with({ value: __ }, (x) => 1)
        .run();

      match(input)
        .exhaustive()
        .with({ value: 'a' }, (x) => 1)
        .with({ value: 'b' }, (x) => 1)
        .run();
```

I guess it can be solved by using `ShallowExclude<i, MatchedValue<i, p>>`  instead of `Exclude<i, MatchedValue<i, p>>`, where `ShallowExclude` might be something like this:

```typescript
export type ShallowExclude<o, e> =
  o extends any[] ? (e extends any[] ? ShallowExcludeArray<o, e> : o)
  : o extends object ? ShallowExcludeObj<o, e>
  : Exclude<o, e>;

type ShallowExcludeMap<o, e> = {
  [k in keyof o]: k extends keyof e ? ShallowExclude<o[k], e[k]> : o[k]
}

type ShallowExcludeObj<o, e>
  = ShallowExcludeMap<o, e> extends {[key: string]: never}
  ? never
  : ShallowExcludeMap<o, e>;
```

However the array matching case seems a bit more complicated... so I'm not sure how to implement `ShallowExcludeArray` (i've tried several things unsuccessfuly... i can always find an edge case scenario :'(  which breaks my attempts)

Maybe you have an idea ? I could try to investigate that a bit further when I have some time.

(disclaimer: my suggestion is poorly tested, and it yet adds some non negligible CPU load when type checking - my laptop fan literaly took off when playing with tests !)

oguimbal:
The general case is a tough nut to crack, and requires some time. But dont you think that it could be a good thing to publish a first version that only supports string-ish exhaustive matching ? That's I think the vast majoritiy of the use cases.

(I know, you'd have to be sure that this `.exhaustive()` thing will be the final api... but dont you think this will ?)

As long as I'm concerned, this prevents me to use your lib at all (I dont have any use for non exhaustive matches on string-related things), which is, I think, sad :(   (and I dont want to publish a fork that might not match your final api).

AndrewMorsillo:
I agree with @oguimba,  string-ish exhaustive matching likely covers 95% of cases. At least in all of my apps it would. I'd propose a separate top level export for the API like `matchExhaustive(...)` instead of `match(...).exhaustive()`. The `match(...).exhaustive()` would be too easy to accidentally forget and let slip in a non-exhaustive check bug.

gvergnaud:
I think the issue we face with this exhaustive matching proposal is not that much about string literals vs other literal types, but more about making `Exclude` work on any level of nesting in your input data structure. Sadly, `Exclude` only works with top level union types like `{t: 'a'} | {t: 'b'}`, but doesn't work for `{t: 'a' | 'b'}`. `Exclude<{t: 'a'} | {t: 'b'}, {t: 'a'}>` will return `{t: 'b'}`, but `Exclude<{t: 'a' | 'b'}, {t: 'a'}>` will return `{t: 'a' | 'b'}`. 

We also need to take into account the case were a data structure contains several unions; like `type Input = { type: 'a' | 'b', value: string | number }`. We would like to make sure that `OurExclude<Input, { type: 'a', value: string }>` keeps the `{ type: 'a', value: number }` and the `{ type: 'b', value: string }` cases. 

I thought a bit about that and I think the best way we can solve this issue would be to have a `DistributeUnion<T>` type function that would turn types like `{ type: 'a' | 'b', value: string | number }`  into `{ type: 'a', value: string } | { type: 'a', value: number } | { type: 'b', value: string } | { type: 'b', value: number }`. Then we could simply use Exclude on it to get back the correct exclusion:  `type OurExclude<a, b> = Exclude<DistributeUnion<a>, b>` . I tried to write this `DistributeUnion` type but I didn't succeed :( I don't see why it wouldn't be possible though. If you find a good solution for it that would be really awesome and we could have an `exhaustive` function that truly delivers what it promises!

oguimbal:
I came with the same conclusion, but I'm not sure it is indeed possible. A good reason I can see why this has not been implemented is combinatorial explosion:  The distributed union you're taling about has a cardinality which is the product of all the cardinalities of all union types that can be found in the object. That can grow pretty quickly, and would be harmful for compile time. 

So if  you dont see which type structure  allows you to implemented this, my 2 cents is that this is on purpose. The designers of TS may have explicitely forbidden such structures which allows combinatorial explosion.

That said, the point about literals was that they're the only case where the distribution you mentioned is pointless: They're already distributed, meaning that it works out of the box with `Exclude`. And switching on a string is most of the use cases. So only implementing this naive edge case is actually implementing almost all of the use cases... it is worth it.

gvergnaud:
> So if you dont see which type structure allows you to implement this, my 2 cents is that this is on purpose. The designers of TS may have explicitely forbidden such structures which allows combinatorial explosion.

I don't think so, the cardinality of the resulting type actually depends on your input, so it's bad for compiler performance only if your input type is very complex. There are other ways to compute very big unions in TS, I don't think they would forbid this on purpose. I think it should be possible to write this type, but it's pretty hard.

> That said, the point about literals was that they're the only case where the distribution you mentioned is pointless: They're already distributed, meaning that it works out of the box with Exclude. And switching on a string is most of the use cases. So only implementing this naive edge case is actually implementing almost all of the use cases... it is worth it.

I'm not sure I understand this. Nested string unions aren't distributed, as you pointed out in one of your previous comments the current naive implementation doesn't work for types like `{ type: 'a' | 'b' }`

oguimbal:
Yes I know, my earlier proposition doesnt work (that was only exploratory).

I was only suggesting here to begin supporting .exhaustive()  if your input type extends string. 
This is a very naive use case,  where the resulting matching is just an equivalent to an exhaustive switch on a string, then. But with a nicer syntax:

```typescript
const result = match(myUnion)
   .exhaustive()
   .with('a', () => 1)
   .with('b', () => 2)
   .with('c', () => 3)
   .run()
```

Instead of this, which I find not nice to read, and does not guarantee the assignation of "result" at first sight:
```typescript
let result: number:
switch (myUnion) {
     case 'a':
         result = 1;
         break;
     case 'b':
          [...]
      default:
          let check: never = myUnion; // something that type-checks exhaustiveness
         throw whatever;
}

**Comment by oguimbal** [comment]
As for the combinatorial explosion, what would `DistributeUnion<X>` looks like for the the following X:

```typescript
interface X {
    type: 'a' | 'b';
    child: X;
}
```

🤷‍♂️

gvergnaud:
Yeah I agree this is pretty valuable to have some kind of naive `exhaustive` matching. Other cases like `Exclude<Option<string>, { kind: 'none' }>` would also works out of the box though, do you think we should not support them because if the shortcomings of Exclude when facing structures containing union types and instead only permit literal patterns?

Another way to solve this would be to have a good documentation section explain why we don't support structures containing unions, and suggesting

gvergnaud:
> As for the combinatorial explosion, what would DistributeUnion<X> looks like for the the following X:
Good point I didn't think about recursive structures, I thin you would get the infamous `Type is too deep` compilation error, but that would be fine because it would be on the user side. You can already get this kind of error by using a very very deep data structure as input type to `match`.

Yeah I agree this is pretty valuable to have some kind of naive exhaustive matching. Other cases like Exclude<Option<string>, { kind: 'none' }> would also works out of the box though, do you think we should not support them because if the shortcomings of Exclude when facing structures containing union types and instead only permit literal patterns?

Another way to solve this would be to have a good documentation section explain why we don't support structures containing unions, and suggesting to rewrite your type as a flat union instead of nested one

gvergnaud:
👋 

I Spent some more time on this and I think I found a way to make it work for all cases. I used the strategy of implementing a `DistributeUnions<U>` type described in [this comment](https://github.com/gvergnaud/ts-pattern/pull/7#issuecomment-739799877). 
This is still very much WIP, but @oguimbal feel free to give a review and see if it meets your needs

Cheers

src/index.ts
```diff
@@ -118,6 +124,17 @@ const builder = <a, b>(
     }
     return entry.handler(value, entry.select(value));
   },
+
+  /**
+   * ### exhaustiveMatch
+   * Entry point to create pattern matching code branches. It returns an
+   * empty Match case.
+   */
+  exhaustive: (): ExhaustiveMatch<DistributeUnions<a>, b> =>
```
gvergnaud, line 133: This should only be allowed right after calling `match(...)`. It doesn't make sense if you already appended a `.with()`. Maybe I we should expose this as another top level `exhaustiveMatch(...)` function, as @AndrewMorsillo suggested? that might make more sense

src/types/DistributeUnions.ts
```diff
@@ -0,0 +1,155 @@
...
+        {
+          [k in Cast<head, PropertyKey>]: Update<
+            SafeGet<data, k, {}>,
+            value,
+            Cast<tail, PropertyKey[]>
+          >;
+        }
+  : value;
+
+export type DistributeUnions<a> = IsAny<a> extends true
```
gvergnaud, line 153: Todo: Add a comment about what is going on here

src/types/FindSelected.ts
```diff
@@ -0,0 +1,12 @@
+import type { SelectPattern } from './Pattern';
+import type { ValueOf } from './helpers';
+
+export type FindSelected<a, b> = b extends SelectPattern<infer Key>
+  ? { [k in Key]: a }
+  : [a, b] extends [(infer aa)[], [infer p]]
+  ? FindSelected<aa, p> extends infer selected
```
gvergnaud, line 7: In this PR, I'm using top level type recursion and variadic tuples, which are only supported since TS 4.1 and TS 4 respectively. It makes this a breaking change because people using earlier versions of TS will have to upgrade.

src/types/helpers.ts
```diff
@@ -44,3 +44,84 @@ export type UnionToIntersection<U> = (
...
+type NeverKeys<o> = ValueOf<
+  {
+    [k in keyof o]: [o[k]] extends [never] ? k : never;
+  }
+>;
+
+type RemoveNeverKeys<o> = Omit<o, NeverKeys<o>>;
+
+export type ExcludeUnion<a> = IsUnion<a> extends true
+  ? never
```
gvergnaud, line 78: I don't think this is used, should remove

oguimbal:
@gvergnaud  Nice, good job. I suppose I was wrong thinking that the TS compiler team made sure that combinatorial explosion was not possible.

Seems like heavy artilery at play, here.

At first sight, I'd only have two concerns with this thing

1) I suppose you already know it, but typing will fail on recursive structures, though.

image

2) All elegance considerations taken apart, I'm a bit concerned about the performance of this thing... for instance, the below type fails after freezing vscode typechecking for 15seconds (because of combinatorial explosion).
... and i think it's not that uncommon to have types with this level of ... dont know how to call that... unionization ?
For instance, I think that even if types had no recursion, they'd fail for [the parser](https://github.com/oguimbal/pgsql-ast-parser) of [pg-mem](https://github.com/oguimbal/pg-mem) (a lib I maintain) ... which types are [here](https://github.com/oguimbal/pgsql-ast-parser/blob/master/src/syntax/ast.ts)

image


So while your distributed union is seducing, I'm worried it might have a very strong impact on the usability of your lib. What do you think ?

src/index.ts
oguimbal: +1, I like the idea.

Maybe just an `exhaustive()`  top level function ? Or `ematch()` ? (`exhaustiveMatch` sounds a bit cumbersome to me for a name that will be heavily used, but I guess its a matter of opinion 😄)

gvergnaud:
I know the type is pretty heavy but I think correctness is more important in this case. I don't want to advertise exhaustive checking when the lib is actually unable to understand nested union structures. It's probably possible to make it more performant over the long run, but I think we should start with a solution that is at least correct. 

Note that this API is purely additive/opt-in,  which mean you can always fallback to non-exhaustive checking if you are more concerned about performance and less about correctness. 

about your first point, I don't intend to make it work for recursive inputs as we discussed a few weeks ago

oguimbal:
Yup, I remember. That's why I mentioned "I suppose you already know".

I get your point. I was just pointing out the limits of this as it is, in the spirit of being constructively critic.

Anyway, limits or no limits, it's  up to you in the end.

gvergnaud:
>  in the spirit of being constructively critic
And it's very much appreciated :) 

Something we could do is to only flatten unions on up to 2 or 3 level. Making sure we are not going through the whole structure might make sense in some cases. I'm not aware of any way to profile the performance of TS's type checking but that would be very valuable in this case, are you aware of any way to do that?

oguimbal:
> Something we could do is to only flatten unions on up to 2 or 3 level.

Yes, that sound like a pragmatic solution. Which could also solve nested types problem.

> are you aware of any way to do that?

Nope, never done that. Maybe [this](https://github.com/microsoft/TypeScript/wiki/Performance#profiling-the-compiler) will help ? That said, it might be hard to find the actual incriminated type using that.

gvergnaud:
@oguimbal If you are interested, I fixed some of the perf issue we were facing with `DistributeUnions` in the latest release (2.1.2). 
On nested objects, perfs were terrible because the type of pattern was computed from the output of `DistributeUnions`, which could be a huge union of objects, and a  `Pattern` had to be computed for every case. The new versions only use the distributed input for exclusions, but the pattern type is computed from the original input. 

I hope that helps :)

oguimbal:
@gvergnaud I'll have a deeper look when I have time out of curiosity, but if it's not `O(big)` anymore, that's pretty sweet 😁 !

Issue #29, opened 2021-05-01
`__.empty/none` pattern


m-rutter:
**Is your feature request related to a problem? Please describe.**
Sometimes something can be a union of `null | undefined | T`. This comes up quite often in graphql projects where `null` and `undefined` are basically interchangable. If you use something like graphql-codgen then the codegen interfaces and types will be littered with stuff like this:

```ts
type Maybe<T> = null | T;

type SomeGraphqlType = {
  __typename: "SomeGraphqlType",
  id: Scalars["ID"],
  foo?: Maybe<SomeOtherGraphqlType | YetAnotherGraphqlType> 
}

type SomeOtherGraphqlType = {
  __typename: "SomeOtherGraphqlType"
}

type YetAnotherGraphqlType= {
  __typename: "YetAnotherGraphqlType"
}
```

Sometimes these nullable unions can be quite large, and so you just want to check if its null/undefined and then maybe later match on that union in some other code branch.

Right now you can do `match(x).with(not(undefined), not(null), () => {})`, which works find, but sometimes that can be a bit laborious if what you want to match on in a little more complex than the above example, such as when you want to match against multiple nullable fields at lots of different depths.

**Describe the solution you'd like**
A `__.empty` or `__.none` pattern that matches `null | undefined`. Allowing a user to write this:
```ts
match(x).with({y: not(__.empty)}, ({y}) => {
  // y is not `null | undefined`
});
```

**Describe alternatives you've considered**
Alternatives are:
Use a `when` with a type guard that takes a `null | undefined | T` and returns `x is T`
Provide multiple patterns, one of which is `not(undefined)` and the other which is `not(null)`.

**Additional context**
https://www.graphql-code-generator.com/

gvergnaud:
👋 I think this should be doable in user land using `when`:

```ts
const nullable = when(
  (x: unknown): x is null | undefined => x === null || x === undefined
);

match<{ str: string } | null>({ str: "hello" })
  .with(not(nullable), ({ str }) => str)
  .with(nullable, () => "")
  .exhaustive();
``` 

However I just noticed a bug with `not`. It seems that it isn't able to correctly invert the type of a `when` pattern containing a type guard so currently when you do `not(nullable)` on some input of type `string | null` the type isn't correctly refined to `string` and instead you get `string | null`. I'll try to fix this in the next patch release

gvergnaud:
[...]

PR #35, opened 2021-06-22
Unit Patterns

m-rutter:
So I've brought this up as desirable in #29 and I didn't detect too much enthusiasm, but I thought to make a PR anyway. 

This PR adds a new wildcard pattern called `__.unit`. Its purpose is to match against both of javascript's unit types, `null` and `undefined`. As they are unit types `undefined` and `null` are their own wildcards, but at least in my codebases its very common for them to appear in a union together. For example in graphql-codgen output its very very common to have codegen like this:

```ts
interface Foo {
  __typename: "Foo";
  value?: string | null;
}
```

These kinds of types make it very annoying to match against, esp when they are more complicated than this example and occur in large discriminated unions. For my purposes `undefined` and `null` are equivalent, and having a built in wildcard for both of them could simplify a lot of my patterns.

It seemed quite easy to implement, but I'm sure I haven't got enough tests (esp of the type level variety). But I would like some feedback about whether you find this addition desirable or problematic.

### Example usage:

```ts
import { match, not, __ } from 'ts-pattern';

interface Foo {
  __typename: "Foo";
  value?: string | null;
}

declare const foo: Foo;

match(foo)
  .with({value: __.unit, () => {
    console.log('is a unit type');
  })
  .with({value: not(__.unit)}, ({value}) => {
    console.log(`is a string, ${value.toLowerCase()}`)
  })
  .exhaustive()
```

### DIFF
--- a/src/types/InvertPattern.ts
+++ b/src/types/InvertPattern.ts
@@ -18,6 +18,8 @@ export type InvertPattern<p> = p extends typeof __.number
   ? string
   : p extends typeof __.boolean
   ? boolean
+  : p extends typeof __.unit
+  ? null | undefined
   : p extends NamedSelectPattern<any> | AnonymousSelectPattern | typeof __
   ? unknown
   : p extends GuardPattern<infer p1, infer p2>
@@ -68,6 +70,8 @@ export type InvertPatternForExclude<p, i> = p extends NotPattern<infer p1>
   ? string
   : p extends typeof __.boolean
   ? boolean
+  : p extends typeof __.unit
+  ? null | undefined
   : p extends NamedSelectPattern<any> | AnonymousSelectPattern | typeof __
   ? unknown
   : p extends GuardPattern<any, infer p1>
--- a/README.md
+++ b/README.md
@@ -666,10 +666,12 @@ console.log(output);
 The `__.unit` pattern will match any value of type `null` or `undefined`.
 
 You will **not often need this wildcard** as ordinarily `null` and `undefined`
-are their own wildcards. However, sometimes `null` and `undefined`
-appear in a union together and you may want to treat them as equivalent. This is
-not the cases in many contexts, but if they do appear together in a union and you
-do want to treat them as equivalent then this may come in handy.
+are their own wildcards. 
+
+However, sometimes `null` and `undefined` appear in a union together 
+(e.g. `null | undefined | string`) and you may want to treat them as equivalent.
+This is often not the case in many contexts. However, if they do appear together 
+in a union and you do want to treat them as equivalent then this may come in handy.
 
 ```ts
 import { match, __ } from 'ts-pattern';
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ yarn add ts-pattern
     - [`__.string` wildcard](#__string-wildcard)
     - [`__.number` wildcard](#__number-wildcard)
     - [`__.boolean` wildcard](#__boolean-wildcard)
+    - [`__.unit` wildcard](#__unit-wildcard)
     - [Objects](#objects)
     - [Lists (arrays)](#lists-arrays)
     - [Tuples (arrays)](#tuples-arrays)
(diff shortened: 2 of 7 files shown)


m-rutter:
I've added docs for it in the `README.md` file highlighting that `null` and `undefined` are their own wildcards and the narrow use case for the unit pattern.

src/types/InvertPattern.ts
```diff
@@ -18,6 +18,8 @@ export type InvertPattern<p> = p extends typeof __.number
   ? string
   : p extends typeof __.boolean
   ? boolean
+  : p extends typeof __.unit
+  ? null | undefined
```
gvergnaud, line 22: I'm a bit afraid by the compile time perf implication of this change. I know it's just another case, but I think we should try to remove cases instead of adding new ones

gvergnaud (commented):
Hey! Nice to see that you are interested in contributing to this project!

In the related issue https://github.com/gvergnaud/ts-pattern/issues/29 I suggested you could create your own `nullable` pattern on your side for this usecase: 

```ts
const nullable = when(
  (x: unknown): x is null | undefined => x === null || x === undefined
);
```

From a DX perspective, do you see any difference between doing

 ```ts
 .with(nullable, () => ...)
  .with({ x: not(nullable) }, () => ...)
```

instead of using the `__.unit` pattern you implemented. Is there some limitations with the type inference when using `when` I don't know about?

I was actually thinking about going in the opposite direction. Remove `__.string`, `__.boolean` and `__.number` from the library, and instead define and export them like this:

```ts
export const number = when(isNumber)
export const boolean = when(isBoolean)
export const string = when(isString)
export const __ = when(isUnknown)
```

in this context it would be very easy to add the `nullable` (or `unit`, even though I'm not sold on the name) pattern to the list.

The benefit of this approach would be to simplify the `Pattern` type by removing special wildcard cases, which could make the compilation a little bit faster.

What do you think?

m-rutter:
Yep, I'm definately interesting in getting involved a bit. I completely understand that `unit` might be a contentious name.

So the motivating example for me what this:
```ts
import {match, when, not} from 'ts-pattern';

// this is a type guard we have in our project that we use widely 
export function notEmpty<TValue>(value: TValue | null | undefined): value is TValue {
  return value !== null && value !== undefined;
}

// this is a type guard I created to deal with the issue with `when` - see below
export function isEmpty<TValue>(value: TValue | null | undefined): value is null | undefined {
  return value === null || value === undefined;
}

type Maybe<T> = T | null;

interface Foo {
    bar?: Maybe<Bar>
    barFallback: string;
}

interface Bar {
  baz: string
}

declare const foo: Foo;

if (notEmpty(foo.bar)) {
  const str: string = foo.bar.baz // works here

}

const value = match(foo)
  .with({bar: when(notEmpty)}, (foo) => {
    const str: string = foo.bar.baz // type error, `when` does not narrow the type
  })
  .otherwise(() => foo.barFallback);


const value2  = match(foo).with({bar: not(when(isEmpty))}, ({bar}) => {
    const str: string = bar.baz // for some reason inverting `isEmpty` does work
  })
  .otherwise(() => foo.barFallback);
```

[playground link](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbxAQxgYwBYBo4HcMCmAdjkRDAL5wBmUEIcA5DAM4C0YqMBURjA3AChhBAB6RYNAK5E0MYBCJwyMAKLgYATwA8AFQBqyADZSCAPgAUAN2OmAXHAO2CcAD7KpRo27gyAJgTUwEQEfgCUDjYmLsAsjobRiIJwcFAEMFK8cFGmcACEALwFHl5wAGRl2c75Rb5EAUEhfkIUIuLQ8NQycgpKsepgWnoJppY5BA5Oie5Ent7u-oHBoRFVibEl83UNy35JKWkZWeNwRcWzpa7uJ2fbS00twoJaYC4AssiaAEYEemanjh8FyMQmEwW4UGoyDQLgAYhAIPsUnAvsgoAB+Bwfb6-ABCaLMyWRqKgsOMRlRaAA1g4WDAoMEAOaPQTgnhQmFwfFQJGogBetPpTMErUEATQRjRLjQijpNARDnhEFBwGocAsKgGWgs1ARADoSWEwkiZUQ5XSoIKGURGQDdRADWjHXzhKLTXKbnAUOgMDqEWEiXrcMAYL6ECSHPhiBryFrNGEKDg-RBjQV-ggiSl3fALVamXb9STnXAAPQluAvFw8OhQHAAAyjRDrcD8EAIcRUyjRdFwFcIFc0ryJFADKT15EIUGDLAIFgsqf+9sdpPJlKpYVBgmza1MACYUsVvZhk2EgyGwxHlOQLI2LP0NPGE0nw2iR6d05m4Nvc3ALfnikW-KluWuo8iw9AuGkyDgX0RBWDw8g2nAdb3oMmjNq27Z4NAVLDqOcDjqGPDTrO87vvKDokmSXhrhughAA)

Essentially I was dissatisfied by the behaviour of `when` in this case, and I struggled to explain to others why one of our most basic type guard failed to narrow the type here, and yet inverting the guard seems to work.

It was also annoying that I could not express the equivalence of `null` and `undefined` outside of a when block when most of them I treat them as equivalent.

```ts
const value = match(foo)
  .with({bar: not(undefined)}, {bar: not(null)}, (foo) => {
    const str: string = foo.bar.baz // as expected this does not narrow the type
  })
  .otherwise(() => foo.barFallback);
```

and this is all fixed with `__.unit`

```ts
const value = match(foo)
  .with({bar: not(__.unit)}, (foo) => {
    const str: string = foo.bar.baz
  })
  .otherwise(() => foo.barFallback);
```

I agree that your solution in #29 works, but I suppose what I wanted was something of a more first class solution from the library that didn't require speical explanation to others. That is my main concern - that I eliminate an edge case that I feel I need to explain away, and I don't really want to say that to be able to effectively use this library in our codebase you need to import this seperate utlity that isn't part of the library.

This seems very good:
```ts
export const number = when(isNumber)
export const boolean = when(isBoolean)
export const string = when(isString)
export const __ = when(isUnknown)
```

src/types/InvertPattern.ts
m-rutter: I was worried about this.

m-rutter:
I've opened a new PR reflecting your feedback in #36

m-rutter:
Closing this as new approach in #36

PR #37, opened 2021-07-05
Pass matched value to otherwise

oguimbal:
Hi !

Wouldnt it be relevant to pass the matched value to the `.otherwise()` handler ?

before:
```typescript
const value = computeValue();
const result = match(value)
        .when(a, d => d * 2)
        .when(b, d => d / 2)
        .otherwise(() => value + 2);
```

after:
```typescript
const result = match(computeValue())
        .when(a, d => d * 2)
        .when(b, d => d / 2)
        .otherwise(d => d + 2);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,10 +354,10 @@ You can use it at the top level, or inside your pattern.
 
 Note that exhaustive pattern matching is **optional**. It comes with the trade-off of having **longer compilation times** because the type checker has more work to do.
 
-Alternatively you can use `.otherwise()`, which take an handler returning a default value. `.otherwise(handler)` is equivalent to `.with(__, handler).exhaustive()`.
+Alternatively you can use `.otherwise()`, which take an handler which is given as argument the value that has not mached other clauses, and returns a value. `.otherwise(handler)` is equivalent to `.with(__, handler).exhaustive()`.
 
 ```ts
-  .otherwise(() => state);
+  .otherwise(value => state);
 ```
 
 If you don't want to use `.exhaustive()` and also don't want to provide a default value with `.otherwise()`, you can use `.run()` instead:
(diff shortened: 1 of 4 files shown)


m-rutter:
What is the advantage of this Vs just using the value in the outer scope? Isn't it the same value?

oguimbal:
That's just for convenience... it allows to avoid a declaration, making the overall style a bit more functional.

This is especially usefull when using ts-pattern "inline", like I do in react code:

```tsx

const someAtom = atom(...);

const Component = () => <div>match(useRecoilValue(someAtom))
                                                   .when('loading', () => <div>Loading</div>)
                                                   .when(..., () => ...)
                                                   .otherwise(d => <div>Unhandled case: {d}</div>)
                                             </div>
```

tests/when.test.ts
```diff
@@ -197,4 +197,12 @@ describe('when', () => {
...
         .run();
   });
+
+
+  it('should pass matched value to otherwise', () => {
+    const result = match(42)
+      .with(51, d => d)
+      .otherwise(d => d);
+    expect(result).toBe(42);
+  })
```
gvergnaud, line 207: Since this is unrelated to the `match(...).when` method, I think it would be better to put this test in a separate `otherwise.test.ts` file

gvergnaud (commented):
This makes sense and the diff looks ok. Could you please update the documentation in the README in this PR as well before we merge it?

oguimbal:
Hi !

I dont know if that's what you had in mind for the README, but it is done (and I created a brand new `otherwise.test.ts` file)

gvergnaud:
Hey! I meant editing the API reference section in the readme, but don't worry about I'll merge this and take care of it.

PR #38, opened 2021-07-05
Exhaustive fallback

oguimbal:
Hi !

Consider the following dummy case:

```typescript
const value = fs.readFileSync('some-file', 'utf8') as AorB;

const value = match(value)
    .with('a', () => 'something')
    .with('b', () => 'other thing')
    .exhaustive();
```

Currently, `.exhaustive()`  checks that you handled every possible type, but does not provide any mechanism other than throwing an exception if the actual runtime value does not conform to this type. Nor any way to customize the thrown exception.

Those are things i'd like to be able to do:

```typescript
// throw a custom exception
.exhaustive(() => {throw new Error('Your configuration file is corrupted')});

// return a default value
.exhaustive(v => {
     console.warn('The config file is corrupted, found value : ' + v);
     return defaultConfig;
});
```

The main difference here compared to `.otherwise()` it that it still checks that every valid case declared in types has been covered, while providing the graceful fallback that `.otherwise()` provides.

(I know I could count on libs like Joi to handle data validation, but that I think that would be largely overkill and redundant)

What do you think ?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -137,7 +137,21 @@ const builder = <a, b>(
         ])
       ).run(),
 
-    exhaustive: () => run(),
+    exhaustive: <c>(handler?: (notMatchedValue: unknown) => c) => {
+      if (!handler) {
+        return run();
+      }
+      return builder<a, PickReturnValue<b, c>>(
+        value,
+        cases.concat([
+          {
+            test: () => true,
+            handler,
+            select: (value) => value,
+          },
+        ])
+      ).run();
+    },
 
     run,
   };
--- a/src/types/Match.ts
+++ b/src/types/Match.ts
@@ -137,7 +137,7 @@ export type Match<
    * */
   exhaustive: DeepExcludeAll<i, patternValueTuples> extends infer remainingCases
     ? [remainingCases] extends [never]
-      ? () => PickReturnValue<o, inferredOutput>
+      ? ExhaustiveResult<o, inferredOutput>
       : NonExhaustiveError<remainingCases>
     : never;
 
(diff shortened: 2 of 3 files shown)


m-rutter:
I kinda think that data at I/O boundaries is exactly when you would either validate your data (joi, yup, io-ts, zod, runtypes, etc...) and/or use `otherwise`. Trying to use `.exhaustive` at an I/O boundary and not expecting an error seems kinda nonsensical at least where I'm standing.

paul-sachs:
Speaking as just an observer, i also like to use switch statements to catch all possible cases and like that I can rely on doing something like:

```tsx
function assertUnreachable(_x: never): never {
  throw new Error('Unknown component type');
}

switch(input) {
  case 'string1':
    return 'x';
  case 'string2':
    return 'y';
  default:
    throw assertUnreachable(input);
}
```

When I saw this, i assumed we could do that in user land with

```tsx
match(input)
  .with('string1', () => 'x')
  .with('string2', () => 'y')
  .otherwise((other) =>  throw assertUnreachable(other));
```

But i noticed the type of `other` isn't narrowed by preceding `with` operations.

jjhiggz:
Not realizing that this PR was up, I went ahead and built something similar. Curious what y'all think

https://github.com/gvergnaud/ts-pattern/pull/219/files

For me I personally like the idea of a `safeExhaustive` function because it's a bit more clear what the intent of that function is, in other words when I need that safety I want to make sure I have it and don't miss it because I forgot to plug in an argument to `exhaustive`. I also supplied an optional second argument for handling that error so you could plug it into your own logging utilities. 

This has definitely come across as a really really useful feature for me though so I'm extremely happy to see the support for it.

gvergnaud:
I know it's been 3 years, but after being initially opposed to this feature, I now think this can make a lot of sense in some contexts. The reason why I didn't consider this a priority was that it could be emulated by wrapping an exhaustive match expression into a try catch, but I think we should probably embrace the unsoundness of TypeScript's type system, make catching unexpected values possible anyway. 

I also like the idea of introducing this feature as an overload, instead of making it a new method. I pulled out the bits of this PR which are still relevant in v5 in this PR: https://github.com/gvergnaud/ts-pattern/pull/253

I need to check that I can preserve the developer experience and the quality of error messages with an overloaded `exhaustive` function type, but this is likely to land in the next minor release

gvergnaud:
#253 was merged! closing this PR

PR #45, opened 2021-09-06
Add Variant types

gvergnaud:
## Motivation

TS-Pattern's goal is to make it possible to pattern-match on any kind of data structure, which means you can start using it in your project without having to change your internal data format. 

That said, discriminated union types are a bit verbose when compared to data constructors that exist on languages like OCaml, Haskell, ReScript, etc:

In TS

```ts
type Shape =
    |  { tag: "Circle", value: { radius: number } } 
    |  { tag: "Rectangle", value: { width: number, height: number } }
    |  { tag: "Square", value: { size: number } }


const area = (shape: Shape) => 
    match(shape)
        .with({ tag: "Circle", value: select() }, ({ radius }) =>  Math.PI * radius ** 2)
        .with({ tag: "Rectangle", value: select() }, ({ width, height }) => width * height)
        .with({ tag: "Square", value: select() }, ({ size }) => size ** 2)
        .exhaustive()

area({ tag: "Circle", value: { radius: 20 } })
// => 1256.6370614359173
```

We are repeating  `{ tag: "<Name>", value: <Value> }` a lot. In ReScript in comparison, the same code feels let verbose:

```rescript
type Shape =
    | Circle({ radius: number })
    | Rectangle({ width: number, height: number })
    | Square({ size: number } )


let area = (shape: Shape) => 
    switch shape {
        Circle({ radius }) =>  Math.PI * value.radius ** 2
        Rectangle({ value }) => value.width * value.height
        Square({ size }) => size ** 2
    }

area(Circle({ radius: 20 })
// => 1256.6370614359173
```

It looks a lot better because each variant in the union has a matching **constructor**: a function of same name that takes the variant's content and returns a value of this type.

## Solution

Here is a proposal to add an easier way to define Variants and match on them with TS-Pattern:

```ts
import { Variant, implementVariants }  from 'ts-pattern'

type Shape =
    |  Variant<"Circle", { radius: number }>
    |  Variant<"Rectangle", { width: number, height: number }>
    |  Variant<"Square", { size: number }>

const  { Circle, Rectangle, Square } = implementVariants<Shape>()

const area = (shape: Shape) => 
    match(shape)
        .with(Circle({ radius: select() }), (radius) =>  Math.PI * radius ** 2)
        // you can also select everything with `Constructor(select())`:
        .with(Rectangle(select()), ({ width, height }) => width * height)
        .with(Square(select()), ({ size }) => size ** 2)
        .exhaustive()

area(Circle({ radius: 20 }))
// => 1256.6370614359173
```

`Variant<Tag, Content>` is a Generic returning a plain object type with this shape: `{ tag: Tag, value: Content }`.

`implementVariants` takes a union type parameter and generate the **constructors** for each variant of this union. Constructors are simple functions taking the content, and wrapping them in an object with the appropriate tag. 

The objects constructed using a Constructor function have **nothing special**. They are plain object of shape `{ tag, value }`, they don't have any method and don't inherit from any class. They can be serialised and constructed by hand. 

Constructors can also be used to construct a **pattern**. In this case the returned value will be of type `VariantPattern<Tag, Content>`, which is an alias for `Variant<Tag, Pattern<Content>>`. This means you can match with nested constructors as well:

```ts

type Maybe<T> = Variant<'Just', T> | Variant<'Nothing'>;

const { Just, Nothing } = implementVariants<Maybe<unknown>>();

const toString = (maybeShape: Maybe<Shape>) =>
      match(maybeShape)
        .with(
           Nothing(),
           () => 'Nothing'
        )
        .with(
          Just(Circle({ radius: select() })),
          (radius) => `Just Circle { radius: ${radius} }`
        )
        .with(
          Just(Square(select())),
          ({ sideLength }) => `Just Square sideLength: ${sideLength}`
        )
        .with(
          Just(Rectangle(select())),
          ({ x, y }) => `Just Rectangle { x: ${x}, y: ${y} }`
        )
        .with(
          Just(Blob(select())),
          (area) => `Just Blob { area: ${area} }`
        )
        .exhaustive();
        
        
toString(Just(Circle({ radius: 20 })))
// => `Just Circle { radius: 20 }`

toString(Nothing())
// => `Nothing`
```

Using Variants with ts-pattern would be entirely **optional**! `match` will keep working on regular values. This isn't a breaking change and implementing this new feature actually doens't require changing the implementation of `match`.

### Inspiration

This addition is in large parts inspired by https://github.com/practical-fp/union-types. Using a typed Proxy to generate constructors from a union type is a very good idea. Unfortunately the values created with `union-types` aren't compatible with `ts-pattern`, this PR tries to bridge this gap


FredericEspiau:
Wonderful idea but I feel like it's out of scope from what `ts-pattern` is about, which is, as you said it, to work on all data types.

I feel like a better way to do this would be to help people create bridges between `ts-pattern` and the type library of their choice. 

Example: you want to use `zod` ? There is a plugin for that. You want to use https://github.com/paarthenon/variant ? No problem, someone made a plugin for that.

This way you can focus on having the best pattern matching library and delivering a way for people to extend it at their will :)

gvergnaud:
👋 

Thanks for reading the PR and giving your opinion, it's great to have some feedback on this stuff :)

Since TS-Pattern uses predicates to determine if a pattern matches or not, you can already use it with your library of choice pretty easily. For example, you can write a `matchZodSchema` function in user land:

```ts
const matchZodSchema = <S extends z.Schema<any, any>>(schema: S) =>
  when((obj: unknown): obj is z.infer<S> => schema.safeParse(obj).success);
```

and then use it like so:

```ts
const userSchema = z.object({
  name: z.string()
});

const getUserName = (u: unknown) =>
  match(u)
    .with(matchSchema(userSchema), ({ name }) => name)
    .otherwise(() => "Anonymous");
```

https://codesandbox.io/s/example-ts-pattern-zod-h0z2q

I'm not sure we need a "plugin system" for that. A predicate function is enough (but admittedly those predicates can be hard to type correctly).


> Having said that, I'd be using a ts-pattern-specific structure extensively throughout the code base. I really like the library, but I'm hesitant to do that – does that make sense? Am I misunderstanding how this would work in the real world?

I understand the wariness about using TS-Pattern specific data constructor throughout a big codebase, that's why I would like those variants to not be specific to the library. My goal is to make it easier to write discriminated union types by automating the creation of constructors via the `implementVariants` function, but those constructors are just simple function returning a value of a type you defined (using `Variant<YourType>`). The instances aren't special either, they are just plain, serialisable JS objects, which makes them usable in any context.

I decided to open this PR because I feel like most of the variant libraries out there are too complex and not composable enough. From my exploration, it looks like most of them return class instances with methods on them to match on their different cases (something like `maybeUser.match({ some: user => {}, none: () => {} })`), which means you can't nest them, and you can't pass them to parts of your codebase that aren't using the library. That's why I started thinking about a simpler variant library that would essentially be syntactic sugar to define regular discriminated union types and that would play well with ts-pattern.

steveadams:
@gvergnaud Thank you for explaining this for me - I understand much better now. I can see that a `Variant` is less intrusive of a convention than I thought. I definitely misunderstood some parts of how this would be used. By reading through the test cases more closely I think I clued into a few things better as well.

The more I look at it and consider how I'd use it in my own work, I like it quite a bit. Our use case would be the one I mentioned with creating domain variants and then leveraging them as far upstream in our logic as possible such that we can gradually avoid writing code which works on any type of domain yet is intended for a specific type of domain. It creates an astounding surface area for bugs.

This is a sore spot in our code base with many possible solutions, but it's actually why we adopted `ts-pattern`. It's a large component of our solution, and variants seem to fit directly in line with our needs.

What are the issues you currently see with the PR? Do you still want to improve upon it, or are there blocking issues?

eboody:
hello!

I dont understand (yet) how to do pull requests and stuff just yet, but i discovered ts-pattern shortly after discovering Zod and i LOVE it! It makes the typescript experience closer to Rust which im learning now.

I have a couple of thoughts that i'd like to share. Maybe you've thought of them, maybe you havent.

First, regarding variants, i've been able to accomplish a similar thing by actually creating objects of the same shape as the pattern i want to match against and adding 'as const' at the end like so:

```
export const Nothing: Nothing = { is: 'nothing', thing: undefined } as const;
export const Something = { is: 'something' } as const;
```

these two objects above are to accompany these:
```
export interface Nothing {
    readonly is: 'nothing';
    readonly thing: null | undefined;
}
export const Some = <Thing>(thing: Thing) => {
    return { is: 'something', thing } as const;
};
export type Option<Thing> = Some<Thing> | Nothing;
```

With these, I am able to do things like:
```
match(maybeSomething)
.with(Something, ({thing})=>{thing})
.with(Nothing, ()=>"....")
.exhaustive()
```

This way, I get the exhaustiveness check, and the ability to match against variants. I'm kind of curious if something like this is achievable for arbitrary variants, abstracting away the verbosity behind the scenes or something.

I'd also like to take this opportunity to describe how i'm using ts-pattern to replace zod's only feature that i want ts-pattern to have and thats parsing where it tells you whats wrong. 

I didn't spend any time (yet) making this more robust but i'll share it here anyways. What I wanted was to be able to parse an object and have it tell me whats of the wrong type.

The next step is for me to actually construct an object of all of the things that are wrong and what it was expecting rather than doing it one at a time.

```
type PatternObject<T> = { [k in string]: P.Pattern<T> };
const isObject = <T>(obj: any): obj is PatternObject<T> => {
    return typeof obj === 'object' && obj !== null;
}

export const parseObject = <T>(pattern: PatternObject<T>, obj: any): Result<T, string> => {

    for (const key in pattern) {
        if (!isMatching(pattern[key], obj[key])) {

            if (typeof obj[key] === 'object') {
                if (isObject(pattern[key])) {
                    if (!isErr(parseObject(pattern[key] as PatternObject<T>, obj[key]))) {
                        continue;
                    }
                }
            }

            return Err(`Invalid value for key ${key}: ${JSON.stringify(obj[key])}`);
        }
    }
    return Ok(obj as T);
}
```

I think it would be a fun side-project for me to contribute to ts-pattern so if you think im on to something I would be overjoyed!

For completeness i'll include the rest of the types:
```
import { P, isMatching } from 'ts-pattern';

export type Nullish<T> = T | null | undefined;

export type Option<Thing> = Some<Thing> | Nothing;

export interface Nothing {
    readonly is: 'nothing';
    readonly thing: null | undefined;
}
export const Some = <Thing>(thing: Thing) => {
    return { is: 'something', thing } as const;
};

export const Nothing: Nothing = { is: 'nothing', thing: undefined } as const;
export const Something = { is: 'something' } as const;

export type Some<Thing> = { readonly is: 'something'; readonly thing: Thing };


export const isSome = <Thing>(option: Option<Thing>): option is Some<Thing> => {
    return option.is === 'something';
};
export const isErr = <Thing, Error>(result: Result<Thing, Error>): result is Err<Error> => {
    return !result.ok;
};

export type Result<Thing, Error> = Ok<Thing> | Err<Error>;

export type Err<Thing> = { readonly ok: false; readonly data: Thing };
export const Err = <Thing>(thing: Thing) => {
    return { ok: false, data: thing } as const;
};

export type Ok<Thing> = { readonly ok: true; readonly data: Thing };
export const Ok = <Thing>(thing: Thing) => {
    return { ok: true, data: thing } as const;
};

type PatternObject<T> = { [k in string]: P.Pattern<T> };
const isObject = <T>(obj: any): obj is PatternObject<T> => {
    return typeof obj === 'object' && obj !== null;
}

export const parseObject = <T>(pattern: PatternObject<T>, obj: any): Result<T, string> => {

    for (const key in pattern) {
        if (!isMatching(pattern[key], obj[key])) {

            if (typeof obj[key] === 'object') {
                if (isObject(pattern[key])) {
                    if (!isErr(parseObject(pattern[key] as PatternObject<T>, obj[key]))) {
                        continue;
                    }
                }
            }

            return Err(`Invalid value for key ${key}: ${JSON.stringify(obj[key])}`);
        }
    }
    return Ok(obj as T);
}
```

cevr:
Is there interest in making this a sub import so that it doesn't affect bundle size? something like `ts-pattern/variants`. This would be a great addition. 

Or, is this easily done is userland nowadays?

PR #48, opened 2021-09-16
Provide a matcher wildcard for NaN

mhintz:
`__.number` doesn't match `NaN`, but sometimes you might need to. Because `NaN !== NaN`, this PR adds a matcher function for `NaN`: `__.NaN`.


gvergnaud:
Great addition, thank you for opening this PR!

PR #54, opened 2021-10-03
Convert to ES module

### LINKED ISSUE #53: Support Angular by making an ES module
# Problem

When importing `ts-pattern` in a modern Angular app (tested on 12+), angular complains that:

```
Warning: <module> depends on 'ts-pattern'. CommonJS or AMD dependencies can cause optimization bailouts.
For more info see: https://angular.io/guide/build#configuring-commonjs-dependencies
```

# Desired Solution

Create an es-module version of this library.

# Alternatives

I'm aware of no alternatives.

paviad:
Closes #53

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,10 +1,11 @@
 {
   "compilerOptions": {
-    "module": "commonjs",
+    "module": "ES6",
     "declaration": true,
     "target": "es6",
     "strict": true,
-    "outDir": "lib/"
+    "outDir": "lib/",
+    "moduleResolution": "Node"
   },
   "exclude": ["tests/", "lib/"]
 }


m-rutter:
I'm not a maintainer, but this isn't likely to be merged in its current state. Dropping commonjs completely is a major breaking change. Node environments for example won't understand these imports and exports. Historically targetting commonjs was a safe bet because most bundlers will happily take care of things and it works in node, but that is changing with many front end build tools expecting esm modules. 

A real fix would involve building multiple artifacts both supporting commonjs and esm simultaneously using the package.json `"main"` and `"module"` fields to indicate entry points.

paviad:
@m-rutter thanks, I am really not familiar with the different module systems (common, amd, etc) so I appreciate your input.

Do you think you can perhaps prepare a PR that does this right? I would appreciate the chance to learn how to do it right.

stefnotch:
As long as one is happy with targeting the "active" version of Node.js, then using es modules doesn't seem to be a problem in that regard.

However, I guess I should add that Node.js v14, which doesn't support es modules by default, will not reach its end-of-life until 2023-04-30.

gvergnaud:
Closing in favor of #70

Issue #69, opened 2021-12-30
Exhaustive check fails with tuple of tuples


finalclass:
**Describe the bug**
While matching against a tuple that has a tuple and the inner tuple has variable number of elements, the exhaustive check fails.

**Code Sandbox with a minimal reproduction case**
https://codesandbox.io/s/romantic-architecture-39wbm

However this works fine:
https://codesandbox.io/s/aged-field-vrldb

So it can be seen that if in a tuple we have a tuple which size varies then it fails the exhaustive check.

**Versions**
- TypeScript version: 4.5.4
- ts-pattern version: 3.3.4
- environment: not related, it fails compilation

gvergnaud:
👋 Thanks for the report.

That's actually because ts-pattern v3 doesn't support unary tuples (tuples containing a single element, like your `[type: "Login"]` type). those patterns are currently considered ["list" patterns](https://github.com/gvergnaud/ts-pattern#lists-arrays), which means they will match lists of variable length, as long as all elements in the array match the sub-pattern (`"Login"` in your case). 

This is a pretty confusing behavior and this will change in v4. see this related comment https://github.com/gvergnaud/ts-pattern/issues/46#issuecomment-917597868

finalclass:
Thanks for your answer and a for this wonderful library:)
Do you have any roadmap/release date for v4?

gvergnaud:
Nope, I only started working on it during this Christmas holidays. 

I'm still experimenting with ideas at this point and I don't want to broadly communicate about it or commit to a release date until I'm sure what I want to do is technically feasible :) But I'm pretty confident it will be released in 2022

Issue #62, opened 2021-12-08
select in array/tuple/list infers wrong type


hesxenon:
**Describe the bug**
when using `select` in a `.with` clause the inferred type is `Array<T>` instead of `T`

**Code Sandbox with a minimal reproduction case**
https://codesandbox.io/s/practical-moore-ytmpf?file=/src/index.ts

**Versions**
- TypeScript version: 4.4.4
- ts-pattern version: 3.3.3
- environment: chromium 96 / node 17

gvergnaud:
Hey,

This behavior is expected, an array with a single sub pattern in it is actually a "list pattern" which will match arrays in which each item matches the subpattern. That's why the value you selected ends up being a list of values, there is one for each item in the array. For more info, see the doc section about list patterns:  https://github.com/gvergnaud/ts-pattern#lists-arrays

I know this is a bit confusing. List patterns might change in v4 to avoid the confusion with unary tuples.

hesxenon:
ok, makes sense (and I can definitely see a use-case for this) but especially since you mention tuples right below the linked docs I'd expect this to select e.g. only the first value as long as the pattern I'm matching against is a "tuple", i.e. a readonly array.

Do you think it's possible to enable this like
```typescript
.with([0], (zeroes) => {})
```
and
```typescript
.with([0] as const, ([firstZero]) => {})
```
?

gvergnaud:
Unfortunately readonly arrays are only a thing at the type level, so there is no way for the runtime to know whether an array is readonly or not and branch on that information to have different behaviors. I think the best way to solve this is to have a different kind of pattern for lists and reserve the `[subpattern]` pattern for unary tuples.

hesxenon:
hmmmm, I see your point... but the correct type could be inferred with the distinction of `readonly` arrays, right?

As for the runtime, imho it's acceptable to specify _how_ you want to match, i.e. I think additional api like
```typescript
.with([0], match.tuple(([firstZero]) => {}))
```
would be totally fine. And it could even solve the typing problem as well?

I mean I'd think the subpattern use case is probably less common than the tuple matching use case but yeah, just an opinion.

Any thoughts on this?

(P.s.: sorry for the late answer, have stopped getting notifications from github for some reason)

Issue #46, opened 2021-09-11
Array element matching bug on empty arrays


oguimbal:
Hi !

**Describe the bug**
```typescript
import { match, __, not } from 'ts-pattern';

console.log(match([]) // Logs 👉 first
  .with([__], () => 'first')
  .otherwise(() => 'default'));

console.log(match([]) // Logs 👉 first
  .with([not(__.nullish)], () => 'first')
  .otherwise(() => 'default'));
```

Even if the first is a bit ambiguous, I would at least expect the second one NOT to print "first"... 

**Code Sandbox with a minimal reproduction case**

https://codesandbox.io/s/icy-glade-3ty9l?file=/src/index.ts

**Versions**
- TypeScript version: 4.3.4
- ts-pattern version: 3.2.5
- environment: any

gvergnaud:
I agree this behavior is confusing, but it's expected. `[subpattern]` patterns are **list** patterns, matching any list in which all elements match the subpattern, including empty lists. If you want to match empty lists separately a workaround is to add a `.with([], handler)` clause before your `[__]` pattern.

In retrospect, I don't think this was a great API choice because it can be confused with an unary tuple pattern (a pattern matching a list with a single element). I've been thinking about introducing a separate `list(subpattern)` helper that would look like this:

```ts
import { match, list, __ } from 'ts-pattern';

match(someList)
   .with([], () => 'empty list')
   .with([__], () => 'list with a single element')
   .with(list(not(null)), () => 'list with n elements')
   .exhaustive();
```

But this would be a breaking change, which makes it a bit tricky to release.

oguimbal:
It would be tricky indeed 🤔

That said, it would be more intuitive, IMO.  
And it would match Haskells behaviour.

Maybe for ts-pattern v4 🥳 ?

gvergnaud:
Closing since TS-Pattern v4 has been released and fixes this problem. [Release note](https://github.com/gvergnaud/ts-pattern/releases/tag/v4.0.1)

PR #70, opened 2021-12-31
TS-Pattern v4

gvergnaud:
⚠️ As the new features rely heavily on [tail call elimination](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#tailrec-conditional), this version is only compatible with TypeScript version 4.5 or higher.

### Imports

type-specific wildcard patterns have moved from `__.<pattern>` to a new `Pattern` qualified module, also exported as `P` by ts-pattern.

```diff
- import { match, __ } from 'ts-pattern';
+ import { match, Pattern } from 'ts-pattern';


const toString = (value: string | number) =>
  match(value)
-   .with(__.string, (v) => v)
-   .with(__.number, (v) => `${v}`)
+   .with(Pattern.string, (v) => v)
+   .with(Pattern.number, (v) => `${v}`)
    .exhaustive();
```

or

```diff
- import { match, __ } from 'ts-pattern';
+ import { match, P } from 'ts-pattern';


const toString = (value: string | number) =>
  match(value)
-   .with(__.string, (v) => v)
-   .with(__.number, (v) => `${v}`)
+   .with(P.string, (v) => v)
+   .with(P.number, (v) => `${v}`)
    .exhaustive();
```

#### `__`

The top level `__` export was moved to `P._` and `P.any`:

```diff
- import { match, __ } from 'ts-pattern';
+ import { match, P } from 'ts-pattern';


const toString = (value: string | number) =>
  match(value)
-   .with(__, (v) => `${v}`)
+   .with(P._, (v) => `${v}`)
    // OR
+   .with(P.any, (v) => `${v}`)
    .exhaustive();
```

#### `select()`, `not()`, `when()`

Function to create patterns have been moved to the `P` module.

```diff
- import { match, select, not, when } from 'ts-pattern';
+ import { match, P } from 'ts-pattern';


const toString = (value: number) =>
  match(value)
-   .with({ prop: select() }, (v) => `${v}`)
+   .with({ prop: P.select() }, (v) => `${v}`)

-   .with({ prop: not(10) }, (v) => `${v}`)
+   .with({ prop: P.not(10) }, (v) => `${v}`)

-   .with({ prop: when((x) => x < 5) }, (v) => `${v}`)
+   .with({ prop: P.when((x) => x < 5) }, (v) => `${v}`)
    .exhaustive();
```

#### `Pattern` type

the `Pattern` type which used to be exported at the toplevel is now accessible at `P.Pattern`.

```diff
- import { match, Pattern } from 'ts-pattern';
+ import { match, P } from 'ts-pattern';

- const pattern: Pattern<number> = P.when(x => x > 2);
+ const pattern: P.Pattern<number> = P.when(x => x > 2);
```

### list patterns

The syntax for matching on a list of elements with an unknown length has changed from `[subpattern]` to `P.array(subpattern)`.

Example:

```diff
- import { match, __ } from 'ts-pattern';
+ import { match, P } from 'ts-pattern';


const parseUsers = (response: unknown) =>
  match(response)
-   .with({ data: [{ name: __.string }] }, (users) => users)
+   .with({ data: P.array({ name: P.string }) }, (users) => users)
    .otherwise(() => []);
```

Now `[subpattern]` matches arrays with 1 element in them. This is more consistent with native language features, like destructuring assignement and is overall more intuitive. This will resolve [#69](https://github.com/gvergnaud/ts-pattern/issues/69), [#62](https://github.com/gvergnaud/ts-pattern/issues/62) and [#46](https://github.com/gvergnaud/ts-pattern/issues/46).

### NaN

The `__.NaN` pattern has been replaced by simply using the NaN value in the pattern:

```diff
match<number>(NaN)
-   .with(__.NaN, () => "this is not a number")
+   .with(NaN, () => "this is not a number")
    .otherwise((n) => n);
```

## New features

Here is the list of all new features which have been added in TS-Pattern v4.

#### `P.array(pattern)`

To match an array of elements, you can now use `P.array`:

```ts
import { match, P } from 'ts-pattern';

const responsePattern = {
  data: P.array({
    id: P.string,
    post: P.array({
      title: P.string,
      content: P.string,
    }),
  }),
};

fetchSomething().then((value: unknown) =>
  match(value)
    .with(responsePattern, (value) => {
      // value: { data: { id: string, post: { title: string, content: string }[] }[] }
      return value;
    })
    .otherwise(() => {
      throw new Error('unexpected response');
    })
);
```

#### `P.optional(pattern)`

If you want one of the keys of your pattern to be optional, you can now use `P.optional(subpattern)`.

If you `P.select()` something in an optional pattern, it's type will be infered as `T | undefined`.

```ts
import { match, P } from 'ts-pattern';

const doSomethingWithUser = (user: User | Org) =>
  match(user)
    .with(
      {
        type: 'user',
        detail: {
          bio: P.optional(P.string),
          socialLinks: P.optional({
            twitter: P.select(),
          }),
        },
      },
      (twitterLink, value) => {
        // twitterLink: string | undefined
        /**
         *  value.detail: {
         *      bio?: string,
         *      socialLinks?: {
         *          twitter: string
         *      }
         *  }
         **/
      }
    )
    .otherwise(() => {
      throw new Error('unexpected response');
    });
```

### Union & intersection patterns

`P.union(...patterns)` and `P.intersection(...patterns)` combine several patterns into a single one, either by checking that one of them match the input (`p.union`) or all of them match it (`P.intersection`).

#### `P.union(...patterns)`

```ts
type Input =
  | { type: 'a'; value: string }
  | { type: 'b'; value: number }
  | {
      type: 'c';
      value:
        | { type: 'd'; value: boolean }
        | { type: 'e'; value: string[] }
        | { type: 'f'; value: number[] };
    };

const f = (input: Input) =>
  match(input)
    .with(
      { type: P.union('a', 'b') },
      // x: { type: 'a'; value: string } | { type: 'b'; value: number }
      (x) => 'branch 1'
    )
    .with(
      // P.union can take any subpattern:
      {
        type: 'c',
        value: { value: P.union(P.boolean, P.array(P.string)) },
      },
      (x) => 'branch 2' // x.value.value: boolean | string[]
    )
    .with({ type: 'c', value: { type: 'f' } }, () => 'branch 3')
    .exhaustive();
```

#### `P.intersection(...patterns)`

```ts
class A {
  constructor(public foo: 'bar' | 'baz') {}
}

class B {
  constructor(public str: string) {}
}

const f = (input: { prop: A | B }) =>
  match(input)
    .with(
      { prop: P.intersection(P.instanceOf(A), { foo: 'bar' }) },
      // prop: A & { foo: 'bar' }
      ({ prop }) => 'branch 1'
    )
    .with(
      { prop: P.intersection(P.instanceOf(A), { foo: 'baz' }) },
      // prop: A & { foo: 'baz' }
      ({ prop }) => 'branch 2'
    )
    .with(
      { prop: P.instanceOf(B) },
      // prop: B
      ({ prop }) => 'branch 3'
    )
    .exhaustive();
```

### Select with sub pattern

`P.select()` now can take a subpattern and match only what the subpattern matches:

```ts
type Img = { type: 'img'; src: string };
type Text = { type: 'text'; content: string; length: number };
type User = { type: 'user'; username: string };
type Org = { type: 'org'; orgId: number };

const post = (input: { author: User | Org; content: Text | Img }) =>
  match(input)
    .with(
      { author: P.select({ type: 'user' }) },
      // user: User
      (user) => {}
    )
    .with(
      {
        // This also works with named selections
        author: P.select('org', { type: 'org' }),
        content: P.select('text', { type: 'text' }),
      },
      // org: Org, text: Text
      ({ org, text }) => {}
    )
    .otherwise(() => {
      // ...
    });
```

#### `P.infer<typeof pattern>`

TS-Pattern is pretty handy for parsing unknown payloads like HTTP responses. You can write a pattern for the shape you are expecting, and then use `isMatching(pattern, response)` to make sure the response has the correct shape.

One limitation TS-Pattern had in its previous version was that it did not provide a way to get the TypeScript type of the value a given pattern matches. This is what `P.infer<typeof pattern>` does :)

```ts
const postPattern = {
  title: P.string,
  description: P.optional(P.string),
  content: P.string,
  likeCount: P.number,
};

type Post = P.infer<typeof postPattern>;
// Post: { title: string, description?: string, content: string, likeCount: number }

const userPattern = {
  name: P.string,
  postCount: P.number,
  bio: P.optional(P.string),
  posts: P.optional(P.array(postPattern)),
};

type User = P.infer<typeof userPattern>;
// User: { name: string, postCount: number, bio?: string, posts?: Post[]  }

const isUserList = isMatching(P.array(userPattern));

const res = await fetchUsers();

if (isUserList(res)) {
  // res: User
}
```

#### `P.symbol`

`P.symbol` is a wildcard pattern matching any **symbol**.

```ts
match(Symbol('Hello'))
  .with(P.symbol, () => 'this is a symbol!')
  .exhaustive();
```

#### `P.bigint`

`P.bigint` is a wildcard pattern matching any **bigint**.

```ts
match(200n)
  .with(P.bigint, () => 'this is a bigint!')
  .exhaustive();
```


m-rutter:
Very exciting! I quite like the look of the `union` and improved `select` patterns. What is the reasoning behind moving the wildcard patterns out of `__` into a seperate export `P`?

gvergnaud:
Thanks for your feedback!

> Ah, its about consolidating the other patterns I suppose right?

Exactly. This new version introduces a lot of new patterns and it would makes the type of `__` too big and make the IDE hints from the typescript language server less readable.

  I also don't really like the look of `__.number` etc that much anymore. It looks a bit alien in a typescript code base imo.

m-rutter:
Might there be a case for `P.__`, `P._`, `P.wildcard` or `P.nil`?

gvergnaud:
`P.__` exists in this PR (it's the same thing as `__`), but I'm curious what would be `P.wildcard` or `P.nil`

m-rutter:
`wildcard` because that is what other langauges that have first class pattern matching often call `__` when speaking abou it (rust, ocaml, etc...), and `nil` because that seems to be the current term being used in the tc39 proposal (https://github.com/tc39/proposal-pattern-matching#nil-pattern).

I'm quite happy with `P._` or `P.__` personally.

m-rutter:
> __, select, when and not are still exported by the package along with P, but that's mainly to ease the migration from v3 to v4. I didn't choose to exports stuff like number, string etc from the index file because they are very common names and could be confused with the homonym types. I'm curious of your opinion about this :)

This seems fine. Reducing the cluttering of imports is a good. I use `__.<type>` a lot of more than anything else, so that biggest breaking change for me. I personally wouldn't mind if you marked the standalone `select` or `not` functions as deprecated or removed entirely because I'm going to have to refactor the `__` usages anyway.

m-rutter:
@gvergnaud I'm happy to give this RC a try on a large-ish codebase if you think its ready.

gvergnaud:
@m-rutter that'd be awesome! I just publish a new `@next` version with the latest changes, please let me know how it goes :)

m-rutter:
Ok, so I've done most of the migration for a fairly large project. I'll post some more feedback tomorrow (one or two likely bugs), but for now here are some before and afters with `--extendedDiaglostics`:

This is a ~250k sloc project reference within a larger project that uses ts-pattern a fair bit. ~250 different files import from ts-pattern within this project reference. Don't read too much into the overall type check time - its slow for other reasons.

3.5.5:
```
Files:                         4221
Lines of Library:              8984
Lines of Definitions:        532845
Lines of TypeScript:         246477
Lines of JavaScript:              0
Lines of JSON:               104488
Lines of Other:                   0
Nodes of Library:             23566
Nodes of Definitions:       2124795
Nodes of TypeScript:        1091163
Nodes of JavaScript:              0
Nodes of JSON:               253373
Nodes of Other:                   0
Identifiers:                 950506
Symbols:                    2842361
Types:                      1306508
Instantiations:            13708238
Memory used:               2442261K
Assignability cache size:    660407
Identity cache size:          24096
Subtype cache size:          177929
Strict subtype cache size:   103425
I/O Read time:                0.11s
Parse time:                   2.97s
ResolveModule time:           0.42s
ResolveTypeReference time:    0.00s
Program time:                 4.18s
Bind time:                    1.61s
Check time:                  45.36s
printTime time:               5.07s
Emit time:                    5.10s
transformTime time:           4.23s
commentTime time:             0.09s
I/O Write time:               0.16s
Total time:                  56.25s
Done in 57.06s.
```


4.0.1-rc-5:
```
Files:                         4222
Lines of Library:              8984
Lines of Definitions:        533053
Lines of TypeScript:         246476
Lines of JavaScript:              0
Lines of JSON:               104488
Lines of Other:                   0
Nodes of Library:             23566
Nodes of Definitions:       2126613
Nodes of TypeScript:        1090864
Nodes of JavaScript:              0
Nodes of JSON:               253373
Nodes of Other:                   0
Identifiers:                 951073
Symbols:                    3112616
Types:                      1542419
Instantiations:            16701734
Memory used:               2640809K
Assignability cache size:   1119039
Identity cache size:          24215
Subtype cache size:          178197
Strict subtype cache size:    86275
I/O Read time:                0.11s
Parse time:                   2.89s
ResolveModule time:           0.42s
ResolveTypeReference time:    0.00s
Program time:                 4.12s
Bind time:                    1.80s
Check time:                  49.63s
printTime time:               7.31s
Emit time:                    7.34s
transformTime time:           6.42s
commentTime time:             0.10s
I/O Write time:               0.16s
Total time:                  62.88s
Done in 63.70s.
```

gvergnaud:
@m-rutter Thanks for taking a look! I'm really curious about the potential bugs you have found, would you mind sharing them? 

About performance, I noticed the new version was a little bit slower as well. I spent quite a lot of time trying to improve them, but I think I'm approaching the local maxima of this architecture (I hope I'm wrong).

v3 has several flaws which I fixed in this version. To check if the expression is exhaustive, instead of excluding input cases based on patterns sequentially, ts-pattern v3 exclude cases all at once. Excluding all at once doesn't work very well on inputs containing union types in several layers, here is an example with v3: https://codesandbox.io/s/wonderful-kirch-rw9ew?file=/src/index.tsx:0-646

v4 exclude cases sequentially to the input value, which fixes this issue but is a little longer to type-check. If I figure out how to use the typescript trace analyser, I'll give another try at improving type-checking performances

m-rutter:
@gvergnaud Hey, I'm sorry I just had a long week at work.

This is probably a pretty big one: https://github.com/gvergnaud/ts-pattern/issues/74

m-rutter:
Performance wise probably the thing I'm more concerned about is that when I had errors in my patterns the language server got fairly slow. 

I also got quite a few more "Type instantiation is excessively deep and possibly infinite" where previously I had no errors in 3.x.x. I'll come up with some examples and see if you think they are reasonable. 

I will say that none of the performance stuff is enough to dissuade me from using the library. Its just a little bit slower.

m-rutter:
From what I can tell all of the new cases of "Type instantiation is excessively deep and possibly infinite" involve patterns where someone has made a mistake. For example, matching on the pattern `{foo: undefined}` when the input is of type `{foo: string}`, or where the pattern doesn't match the the input at all like having the pattern `{foo: P.string}` but the input is `{bar: P.string}`.

m-rutter:
To be clear, I'm not sure if the above is a big problem or really a problem at all, but its just something I've noticed.

gvergnaud:
Thanks a lot! I'll look into this soon

gvergnaud:
From what I can tell it seems like there is a performance bug with `not` + exhaustiveness checking. Removing the `P.not(...)` patterns from your example seems to fix the "infinitely deep type" error. I'll investigate

m-rutter:
So I did another from scratch migration with rc.7. Its not a before and after of the same base commit from last time. I ran it a few times, and these times seem roughly representative: 

3.5.5:
```
Files:                         4261
Lines of Library:              8984
Lines of Definitions:        533455
Lines of TypeScript:         247348
Lines of JavaScript:              0
Lines of JSON:               104489
Lines of Other:                   0
Nodes of Library:             23566
Nodes of Definitions:       2127837
Nodes of TypeScript:        1098209
Nodes of JavaScript:              0
Nodes of JSON:               253376
Nodes of Other:                   0
Identifiers:                 953238
Symbols:                    2849459
Types:                      1311024
Instantiations:            13802948
Memory used:               2459865K
Assignability cache size:    663871
Identity cache size:          24392
Subtype cache size:          180058
Strict subtype cache size:   103520
I/O Read time:                0.10s
Parse time:                   2.85s
ResolveModule time:           0.39s
ResolveTypeReference time:    0.00s
Program time:                 4.01s
Bind time:                    1.77s
Check time:                  47.66s
printTime time:               5.26s
Emit time:                    5.28s
transformTime time:           4.46s
commentTime time:             0.09s
I/O Write time:               0.13s
Total time:                  58.72s
Done in 59.66s.
Done in 59.85s.
```

4.0.1-rc.8
```
Files:                         4262
Lines of Library:              8984
Lines of Definitions:        533709
Lines of TypeScript:         247347
Lines of JavaScript:              0
Lines of JSON:               104489
Lines of Other:                   0
Nodes of Library:             23566
Nodes of Definitions:       2129760
Nodes of TypeScript:        1098619
Nodes of JavaScript:              0
Nodes of JSON:               253376
Nodes of Other:                   0
Identifiers:                 954146
Symbols:                    3032394
Types:                      1520389
Instantiations:            18772203
Memory used:               2715970K
Assignability cache size:   1070139
Identity cache size:          24535
Subtype cache size:          179961
Strict subtype cache size:    86500
I/O Read time:                0.12s
Parse time:                   2.92s
ResolveModule time:           0.43s
ResolveTypeReference time:    0.00s
Program time:                 4.13s
Bind time:                    1.75s
Check time:                  49.76s
printTime time:               5.37s
Emit time:                    5.39s
transformTime time:           4.56s
commentTime time:             0.09s
I/O Write time:               0.13s
Total time:                  61.03s
Done in 62.00s.
Done in 62.17s.
```

I'm not an expert at reading these, but looks like the gap in type check time has closed but memory consumption is up a bit. Some of the cache stats have grown a lot - no idea if that is a good or bad thing.

m-rutter:
Compared to rc.5 it looks like an overall win.

m-rutter:
Getting some odd errors from esbuild when attempting to bundle ts-pattern:

```
  You can mark the path "ts-pattern" as external to exclude it from the bundle, which will remove
  this error.

✘ [ERROR] Could not resolve "ts-pattern"

      33 │ import { match, P } from "ts-pattern";
         ╵                          ~~~~~~~~~~~~

  The path "." is not currently exported by package "ts-pattern":

    node_modules/ts-pattern/package.json:7:13:
      7 │   "exports": {
        ╵              ^

  None of the conditions provided ("require") match any of the currently active conditions
  ("browser", "default", "import"):

    node_modules/ts-pattern/package.json:8:9:
      8 │     ".": {
        ╵          ^

  Consider using a "require()" call to import this file:

      33 │ import { match, P } from "ts-pattern";
         ╵                          ~~~~~~~~~~~~

  You can mark the path "ts-pattern" as external to exclude it from the bundle, which will remove
  this error.


```

johnknoop:
Regarding list patterns... does this change only apply to properties of objects, like in the example above with the `data` propery, or does it apply to arrays in general, like in this readme example?

https://github.com/gvergnaud/ts-pattern#example-a-state-reducer-with-ts-pattern

If it's the latter, then that's a bit unfortunate, since the `[a, y, z]` pattern syntax is a lot cleaner than `P.Array()`.

m-rutter:
> Regarding list patterns... does this change only apply to properties of objects, like in the example above with the `data` propery, or does it apply to arrays in general, like in this readme example?
> 
> https://github.com/gvergnaud/ts-pattern#example-a-state-reducer-with-ts-pattern
> 
> If it's the latter, then that's a bit unfortunate, since the `[a, y, z]` pattern syntax is a lot cleaner than `P.Array()`.

If I'm understanding your question correctly, then you can be rested assured that the changes in 4.x.x do not remove tuple patterns. 

Previously, if you wanted to match against an array you had to pass a tuple of length 1. This made such patterns ambiguous between wanting to match against a tuple vs an array, and if I recall correctly expecting to match against 1 length tuples produced unexpected results and was not supported. 

The introduction of `P.array` is so that you can disambiguate between array patterns and tuple patterns:

```ts
type Why = [string];

const why: Why; = ["why"]

match(why)
  // pattern for a tuple of length 1 with the value "why" at the 0 index
  .with(["why"], () => {})
  
type WhyArr = Array<"why">

const whyArr: WhyArr = ["why", "why", "why"];

match(whyArr)
  // pattern for an array of "why"
  .with(P.array("why"), () => {});
```

gvergnaud:
I made some more perf tests between master and v4, here are the results:

```md
| task                       | v3      | v4      |
| -------------------------- | ------- | ------- |
| Files:                     | 149     | 163     |
| Lines of Library:          | 26090   | 27248   |
| Lines of Definitions:      | 24860   | 24860   |
| Lines of TypeScript:       | 7690    | 8797    |
| Lines of JavaScript:       | 0       | 0       |
| Lines of JSON:             | 0       | 0       |
| Lines of Other:            | 0       | 0       |
| Nodes of Library:          | 113641  | 117112  |
| Nodes of Definitions:      | 130750  | 130750  |
| Nodes of TypeScript:       | 42236   | 47009   |
| Nodes of JavaScript:       | 0       | 0       |
| Nodes of JSON:             | 0       | 0       |
| Nodes of Other:            | 0       | 0       |
| Identifiers:               | 97614   | 100592  |
| Symbols:                   | 574081  | 578853  |
| Types:                     | 311757  | 283130  |
| Instantiations:            | 1950063 | 2795145 |
| Memory used:               | 544455K | 531928K |
| Assignability cache size:  | 346259  | 370005  |
| Identity cache size:       | 13874   | 14474   |
| Subtype cache size:        | 2091    | 1936    |
| Strict subtype cache size: | 5361    | 765     |
| I/O Read time:             | 0.02s   | 0.02s   |
| Parse time:                | 0.59s   | 0.59s   |
| ResolveModule time:        | 0.03s   | 0.03s   |
| ResolveTypeReference time: | 0.01s   | 0.01s   |
| Program time:              | 0.69s   | 0.69s   |
| Bind time:                 | 0.30s   | 0.31s   |
| Check time:                | 7.83s   | 8.69s   |
| printTime time:            | 0.00s   | 0.00s   |
| Emit time:                 | 0.00s   | 0.00s   |
| Total time:                | 8.82s   | 9.69s   |
```

The check time is a little longer with v4, but it could be coming from the numerous test cases I added on the v4 branch so this could be a red herring. Overall, there doesn't seem to be a significant difference between v3 and v4 in terms of both compilation time or runtime performances.

m-rutter:
Overall I think I'm quite happy with the changes in 4.x.x and I think the slight performance degradation (its obviously improved over the past few RCs) is worthwhile for the improved checks (ts-pattern 4.x.x caught a 2-3 mistakes that were made in 3.5.5) and improved API. 

The migrations have been mostly painless if a bit fiddly at times, though I imagine people with less of a tolerance than me will find the move from `__.string` to `P.string` annoying.

Another migration:

**N.B:** the scope of the project reference has increased a fair amount since last done, and this is on much slower hardware than last time. Below is on 2019 macbook pro, whereas previous migrations were done on much beefer desktop hardware. I might try it on the beefer desktop once I'm home this evening.

TypeScript v4.6.1-rc | v3.3.5 | v4.0.1-rc.8
-- | -- | --
Files: | 4181 | 4182
Lines of Library: | 9294 | 9294
Lines of Definitions: | 300589 | 300841
Lines of TypeScript: | 245132 | 245150
Lines of JavaScript: | 0 | 0
Lines of JSON: | 104488 | 104488
Lines of Other: | 0 | 0
Nodes of Library: | 24093 | 24093
Nodes of Definitions: | 1234846 | 1236795
Nodes of TypeScript: | 2732691 | 2733452
Nodes of JavaScript: | 0 | 0
Nodes of JSON: | 253373 | 253373
Nodes of Other: | 0 | 0
Identifiers: | 923565 | 924652
Symbols: | 3692843 | 3868886
Types: | 1556435 | 1744064
Instantiations: | 13911741 | 15751760
Memory used: | 3031545K | 3104832K
Assignability cache size: | 671650 | 1047569
Identity cache size: | 24143 | 24303
Subtype cache size: | 169958 | 169845
Strict subtype cache size: | 281942 | 264520
I/O Read time: | 0.27s | 0.28s
Parse time: | 3.48s | 3.58s
ResolveModule time: | 0.79s | 0.78s
ResolveTypeReference time: | 0.02s | 0.02s
Program time: | 5.57s | 5.67s
Bind time: | 2.23s | 2.47s
Check time: | 61.94s | 66.15s
printTime time: | 13.13s | 13.26s
Emit time: | 13.16s | 13.30s
Total time: | 82.90s | 87.58s

gvergnaud:
Thanks for running this performance tests, that's really helpful since I don't have a big codebase using TS v4.5+ at my disposal! 

I have another branch which doesn't use tail call recursion to compute the type of `.exhaustive()`, but uses the same technique as in v3.x. It introduces false negatives (considering not exhaustive expressions that actually are), but maybe the perfs would be better and it would justify rolling back this "improvement". Would you be ok to try it on your codebase if I released another rc with this change in it? I'll test on my end as well, but I feel like my sample isn't big enough

m-rutter:
Yeah, feel free to point me at something

m-rutter:
Desktop PC (5900x):

TypeScript v4.6.1-rc | v3.3.5 | v4.0.1-rc.8
-- | -- | --
Files: | 4181 | 4182
Lines of Library: | 9294 | 9294
Lines of Definitions: | 300589 | 300841
Lines of TypeScript: | 245132 | 245150
Lines of JavaScript: | 0 | 0
Lines of JSON: | 104488 | 104488
Lines of Other: | 0 | 0
Nodes of Library: | 24093 | 24093
Nodes of Definitions: | 1234846 | 1236795
Nodes of TypeScript: | 2732691 | 2733452
Nodes of JavaScript: | 0 | 0
Nodes of JSON: | 253373 | 253373
Nodes of Other: | 0 | 0
Identifiers: | 923565 | 924652
Symbols: | 3692843 | 3868886
Types: | 1556435 | 1744064
Instantiations: | 13911741 | 15751760
Memory used: | 2916050K | 3077800K
Assignability cache size: | 671650 | 1047569
Identity cache size: | 24143 | 24303
Subtype cache size: | 169961 | 169845
Strict subtype cache size: | 281949 | 264520
I/O Read time: | 0.09s | 0.11s
Parse time: | 2.82s | 2.84s
ResolveModule time: | 0.36s | 0.38s
ResolveTypeReference time: | 0.01s | 0.01s
Program time: | 3.88s | 3.93s
Bind time: | 1.87s | 1.98s
Check time: | 42.48s | 45.00s
printTime time: | 7.32s | 7.45s
Emit time: | 7.34s | 7.47s
Total time: | 55.58s | 58.38s

m-rutter:
Yeah, might be worth waiting for a certain milestone being reached such as vscode's built in version being >=4.6.x.

m-rutter:
Btw, I recently started using [api-extractor](https://api-extractor.com/) at work for a library to bundle d.ts files and track changes in the pubic API. Thought you might find it interesting.

gvergnaud:
Thanks for trying the RC out @zoontek!

> Did you considered P.any instead of P.__?

I did, yes, but I thought most people already familiar with pattern-matching would expect an underscore as the catch-all pattern because this is what most languages use. `__` looks fine, but I agree `P.__` looks a bit ugly. I expect that `P.any` would look more natural to typescript developer. Maybe we could add `P.any` as an alias to `__`?

> select, when, not, instanceOf and __ are still accessible outside P. Maybe it would be better for consistency to remove these exports.

I kept them on purpose to ease the migration path of people switching from v3 to v4, but I agree it introduces an inconsistency between these patterns and other patterns that aren't exported from the module. I'll think about this. 

Thank you for giving some feedback, that's really helpful!

gvergnaud:
I ended up: 
* Exporting `./pattern` as both `Pattern` and `P`, because I know some people don't like single letter values
* Renaming `P.__` into `P.any`, but keeping the `__` alias as a top level export
* Moving the `Pattern` type to `P.Pattern` or `Pattern.Pattern`

It's released under `ts-pattern@v4.0.1-rc.15`, let me know what you think

zoontek:
> Thanks for trying the RC out @zoontek!
> 
> > Did you considered P.any instead of P.__?
> 
> I did, yes, but I thought most people already familiar with pattern-matching would expect an underscore as the catch-all pattern because this is what most languages use. `__` looks fine, but I agree `P.__` looks a bit ugly. I expect that `P.any` would look more natural to typescript developer. Maybe we could add `P.any` as an alias to `__`?
> 
> > select, when, not, instanceOf and __ are still accessible outside P. Maybe it would be better for consistency to remove these exports.
> 
> I kept them on purpose to ease the migration path of people switching from v3 to v4, but I agree it introduces an inconsistency between these patterns and other patterns that aren't exported from the module. I'll think about this.
> 
> Thank you for giving some feedback, that's really helpful!

`P.any` looks good for me if `any` is not exported directly. Both can co-exists, but I will only use the first one then. Same, I will not use exported methods, maybe `eslint-plugin-import` could help here to forbid the others.

m-rutter:
I'm not terribly keen on the `P.any` in large part because of your original argument that `_` is existing languages do with this as a first class feature. 

This is in fact what the tc39 proposal for pattern matching does:

# Nil pattern
```
match (someArr) {
  when ([_, _, someVal]): ...
}
```

In addition`P.__` is probably better described as `P.unknown` if we want to be pedantic about it.

zoontek:
@m-rutter Then why not `P._` and `Pattern._` ?

`P.unknown` does not make sense to me. Regarding TS types, I agree that we should use `unknown` other `any`, but we say to `ts-pattern` "if it match any value, pick this branch", not "if it match an unknown value, pick this branch".

m-rutter:
>  "if it match any value, pick this branch", not "if it match an unknown value, pick this branch".

I'm being pedantic because `unknown` is the proper top type in typescript, not `any`. Any value is assignable to a variable of type `unknown`. It is also how the `__` guard is implemented:

https://github.com/gvergnaud/ts-pattern/blob/2d2575e1c185e0bcd092c082744300a4ef44c9a1/src/wildcards.ts#L3


`any` acts as neither the proper top nor bottom type in typescript, and effectively turns off the type checker or is interpreted as `never` or `unknown` whichever is more permissive given the position its used.

gvergnaud:
`any` (legitimately) gets a bad rap in TS, so I'm a bit wary of confusing people and giving them the impression that `P.any` is  unsafe even though it's not. `P.unknown` is a bit too long for something so common imo though. I'm starting to like `P._` more and more, for some reason it doesn't look as awkward as `P.__`. What would be awkward though would be to export `_` a the top level because `_` is usually assumed to be `lodash` or `underscore`.

Would `P._` be enough, without a top level `__` value exported?

zoontek:
For me `P._`, without exporter `__` is great. Keeping both would be weird.
Like you, I find that less awkward 
 than `P.__`, and it matches common practice in others languages:

- https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html
- https://rescript-lang.org/docs/manual/latest/pattern-matching-destructuring
- https://docs.scala-lang.org/tour/pattern-matching.html

gvergnaud:
It's a bit noisy in some cases. 
```ts
const f = (input: readonly [1 | 2 | 3, 1 | 2 | 3, 1 | 2 | 3]) =>
        match(input)
          .with([P.not(1), P.not(1), P.not(1)], (x) => 'ok')
          .with([1, P._, P._], () => 'ok')
          .with([P._, 1, P._], () => 'ok')
          .with([P._, P._, 1], () => 'ok')
          .exhaustive();
```
I feel like it's easier to parse with `__`:

```ts
const f = (input: readonly [1 | 2 | 3, 1 | 2 | 3, 1 | 2 | 3]) =>
        match(input)
          .with([P.not(1), P.not(1), P.not(1)], (x) => 'ok')
          .with([1, __, __], () => 'ok')
          .with([__, 1, __], () => 'ok')
          .with([__, __, 1], () => 'ok')
          .exhaustive();
```

I'm thinking we could keep `__`, `P.__` and `P.any` as different aliases of the same thing and see what sticks...

zoontek:
> `P._` is a bit noisy in some cases:
> 
> ```ts
> const f = (input: readonly [1 | 2 | 3, 1 | 2 | 3, 1 | 2 | 3]) =>
>         match(input)
>           .with([P.not(1), P.not(1), P.not(1)], (x) => 'ok')
>           .with([1, P._, P._], () => 'ok')
>           .with([P._, 1, P._], () => 'ok')
>           .with([P._, P._, 1], () => 'ok')
>           .exhaustive();
> ```
> 
> I feel like it's easier to parse with `__`:
> 
> ```ts
> const f = (input: readonly [1 | 2 | 3, 1 | 2 | 3, 1 | 2 | 3]) =>
>         match(input)
>           .with([P.not(1), P.not(1), P.not(1)], (x) => 'ok')
>           .with([1, __, __], () => 'ok')
>           .with([__, 1, __], () => 'ok')
>           .with([__, __, 1], () => 'ok')
>           .exhaustive();
> ```
> 
> I'm thinking we could keep `__`, `P.__` and `P.any` as different aliases of the same thing and see what sticks...

I personally find the first one better, but it's just me 😄

gvergnaud:
I created a poll on twitter: https://twitter.com/GabrielVergnaud/status/1506583038666448899?s=20&t=FBQarq7xy5weyBc7zvwjkg ;)

SandroMaglione:
I think the second underscore in `P.__` does not add much value, it just makes things more verbose.

Also, if other patterns are all based on `P` (e.g. `P.not`) I think is more intuitive to have also the catch all coming from `P`, and not having `__` at all.

That leaves `P._` and `P.any`. I think between this two it is more a matter of personal preference. I also think that since many people that are used to pattern matching have a background in functional programming, `P._` would be more intuitive for most.

zoontek:
+1 for `P._` and `P.any`, without the top-level export. That's also what the twitter poll tends to say 😀

gvergnaud:
I've updated the PR to expose the wildcard pattern as both `P._` and `P.any` and published a new RC. I think that makes sense because:
- It looks like `P.any` is more intuitive to TS/JS devs who don't necessarilly have an experience with languages supporting pattern matching.
- devs who have used pattern matching in other languages prefer `P._` over `__`.
- Removing the top level `__` export also removes the last inconsistency in the way patterns are exported.

src/internals/symbols.ts
```diff
@@ -0,0 +1,26 @@
...
+ * users should not import them
+ * @module
+ * @private
+ * @internal
+ */
+
+export const toExclude = Symbol('@ts-pattern/to-exclude');
+export type toExclude = typeof toExclude;
+
+export const matcher = Symbol('@ts-pattern/matcher');
```
bloodyowl, line 19: To ease interop, would it be possible to make this one `Symbol.matcher ?? Symbol('@ts-pattern/matcher')`?

gvergnaud: This release doesn't include custom matcher protocol support on purpose, and this symbol isn't supposed to be use externally for now, hence the name of the `internals/` folder ;)

gvergnaud: I agree it would be good to make ts-pattern support a kind of custom matcher protocol that would be optionally implemented by othre libraries, without needing a hard dependency on ts-pattern. A "type only" dependency could be also work.

Even though ts-pattern v4 is architected with this goal in mind, a matcher protocol isn't publicly exposed in the intial version and I don't want to tie the release with the support of this protocol, which imo is much harder to get right and to iterate on.

m-rutter: I'm guessing should the tc39 proposal make progress to stage 3/4 (its up for stage 2 at the end of this month) that there might still be space for libraries that can create or expose various objects that implement the matcher protocol.

gvergnaud:
I'm finally merging this PR and releasing the new major version 🎉  The release will be announced next Monday!

PR #78, opened 2022-03-06
Add ESM support

stefnotch:
This attempts to add ESM support without breaking the existing Commonjs stuff. 
However, this is my first time attempting to apply this type of bandaid, so I may have missed a thing or two.

See also: https://github.com/gvergnaud/ts-pattern/pull/54

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,10 +2,17 @@
   "name": "ts-pattern",
   "version": "4.0.1-rc.10",
   "description": " The exhaustive Pattern Matching library for TypeScript.",
-  "main": "lib/index.js",
+  "main": "lib/cjs/index.js",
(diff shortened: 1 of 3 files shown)


gvergnaud:
Hey @stefnotch thanks for your contribution, but I'm actually considering using something like https://github.com/developit/microbundle instead. I did a quick check on my branch and published `ts-pattern@4.0.1-rc.12` with it and it seems to be working pretty well. If you want to try it and tell me if that solves your problem that'd be great!

stefnotch:
I just tried it out, that seems to work like a charm. While Microbundle is new to me, it definitely looks like a solid choice.

That means, this PR can be closed :)

PR #84, opened 2022-03-30
Clarify documentation of .run()

### LINKED ISSUE #83: Document difference between .run() and .exhaustive()
As far as I understand, the difference between `.run()` and `.exhaustive()` is only relevant for type checking, i.e. they behave the same at run time (throwing an error when no match was found).

I'm not sure why you would even want to use `.run()`?
I'm writing this because I was expecting it to behave like `.otherwise(() => undefined)` but based on my observation that is not the case.

I think the documentation of `.run()` should clarify the distinction (and use case, if any).

If you confirm that I have indeed understood how it works, I'd be happy to supply a PR.

Thanks for the great project! <3

philer:
Fix #83

### DIFF
--- a/README.md
+++ b/README.md
@@ -591,7 +591,7 @@ match(...)
   .run()
 ```
 
-Runs the pattern-matching expression and returns its result.
+Runs the pattern-matching expression and returns its result. It throws an error at run time if no match was found, same as `exhaustive()`. However, unlike `.exhaustive()`, exhaustiveness is not check at compile time, meaning the type checker will not verify that all possible cases are covered.
 
 #### Signature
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -591,7 +591,7 @@ match(...)
   .run()
 ```
 
-Runs the pattern-matching expression and returns its result.
+Runs the pattern-matching expression and returns its result. It throws an error at run time if no match was found, same as `exhaustive()`. However, unlike `.exhaustive()`, exhaustiveness is not checked at compile time, meaning the type checker will not verify that all possible cases are covered.
```
gvergnaud, line 594: Awesome, thanks!

gvergnaud (approved):

PR #133, opened 2023-01-08
[chainable] add chainable methods

gvergnaud:


gvergnaud:
merged as part of #139

PR #157, opened 2023-05-05
New patterns: P.startsWith and P.endsWith

knpwrs:
This PR starts two new patterns, `P.startsWith` and `P.endsWith` (I say starts because I haven't written documentation yet). For some reason, `.exhaustive()` causes a type error in `starts-with.test.ts`, but not `ends-with.test.ts`. I'm not sure what's going on yet (hence no documentation) but I'm definitely open to feedback.

### DIFF
--- a/src/patterns.ts
+++ b/src/patterns.ts
@@ -360,6 +360,16 @@ function isInstanceOf<T extends AnyConstructor>(classConstructor: T) {
     val instanceof classConstructor;
 }
 
+function doesStartWith<T extends string>(suffix: T) {
+  return (val: unknown): val is `${T}${string}` =>
+    typeof val === 'string' && val.startsWith(suffix);
+}
+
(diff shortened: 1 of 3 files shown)


gvergnaud:
Hey!

Thanks for the PR. I'm actually working on including these methods in the next major release of TS-Pattern (v5!). They are going to be on the `P.string` namespace:

```ts
const f = (input: string | number) =>
        match(input)
          .with(P.string.startsWith('http'), handle)
          .with(P.string.endsWith('!'), handle)
          .with(P.string.includes('hello'), handle)
          .with(P.string.regex(/^https?:\/\//), handle)
          .exhaustive()
```

knpwrs:
Awesome. Do you happen to know what's going on with the type error I mentioned?

gvergnaud:
@knpwrs no idea, but I included your tests in my PR and they are passing with `P.string.startsWith` :)

Issue #218, opened 2024-01-27
.safeExhaustive


jjhiggz:
**Is your feature request related to a problem? Please describe.**

In my app that I'm working on, I love the developer experience of exhaustive checks, because they force me to handle all of the edge cases that I need. But too often I am encountering scenarios where something doesn't migrate quite right, my runtime type becomes slightly different than my compile time type, and then I get a production error and have to switch from `exhaustive` to `otherwise`.

**Describe the solution you'd like**

It would be extremely awesome if there was a safer alternative to .exhaustive that provided the same developer experience in typescript. To me this would look something like this:

```ts
type MyUnion = "a" | "b"

const a = "" as MyUnion

match(a)
  .with(("a") => "you chose a")
  .with(("b") => "you chose b")
  .safeExhaustive(
     "a" // <= default  if error is thrown by patern match,
     console.error // a callback that allows us to handle the error throw by pattern matching
  )
```

**Describe alternatives you've considered**
Using .otherwise

In a perfect world (I know this is probably a bigger ask), it would be cool if we could create a customized instance of `match` so we wouldn't have to dump in a callback every time or something like that. But I imagine that's a much bigger lift than just building a `.safeExhaustive` method

Liam-Tait:
Similar to https://github.com/gvergnaud/ts-pattern/issues/144

gvergnaud:
v5.7 adds an optional fallback to `.exhaustive` that makes it behave like the `safeExhaustive` method proposed here. Closing

PR #219, opened 2024-01-27
added a safe exhaustiveness check

jjhiggz:
I haven't done much open source, but I thought this might be a really useful function for me personally.

[Here is a link to the issue that I opened](https://github.com/gvergnaud/ts-pattern/issues/218)

### DIFF
--- a/src/match.ts
+++ b/src/match.ts
@@ -126,6 +126,28 @@ class MatchExpression<input, output> {
     );
   }
 
+  safeExhaustive(
+    handler: (value: input) => output,
+    handleErrorMessage?: (input: string) => unknown
+  ): output {
+    if (this.state.matched) return this.state.value;
+
+    let displayedValue;
+    try {
+      displayedValue = JSON.stringify(this.input);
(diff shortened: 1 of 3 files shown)


MatanYadaev:
@gvergnaud What's your opinion about this feature? We need a way to catch both compile-time and runtime errors. Currently `exhaustive` and `otherwise` provide just one of the errors (compile-time and runtime, respectively)

gvergnaud:
I'm still considering it, but I'm not sure what's the best option in terms of API complexity / typesafety yet. In the meantime, this work around doesn't seem too bad:

```ts
try {
  return match(...).with(...).exhaustive()
} catch (e) {
  return defaultValue
}

MatanYadaev:
@gvergnaud it looks great imo. Are you ok with the breaking changes? Don't you prefer a backward compatible change?

gvergnaud:
#253 was merged, adding an optional fallback function to `.exhaustive` which covers the same use case as `.safeExhaustive` proposed here. Closing

PR #234, opened 2024-03-31
Add P.object and P.object.empty

gvergnaud:
## `P.object` predicates

`P.object` is itself a pattern, but also a module containing predicates related to object types.

### `P.object`

`P.object` matches any value assignable to the `object` TypeScript type. This includes all object literals, but also **arrays** and **functions**!

`P.object` does not match primitive types, like strings or numbers.

```ts
import { match, P } from 'ts-pattern';

const fn = (input: unknown) =>
  match(input)
    .with(P.object, () => '✅')
    .otherwise(() => '❌');

console.log(fn({})); // ✅
console.log(fn({ hello: 'world!' })); // ✅
console.log(fn([])); // ✅
console.log(fn(() => {})); // ✅

console.log(fn(1, true, 'hi')); // ❌ ❌ ❌
```

### `P.object.empty()`

`P.object.empty()` matches the empty object `{}`:

```ts
import { isMatching, P } from 'ts-pattern';

console.log(isMatching(P.object.empty(), {})); // true
```

`P.object.empty()` does **not** match empty arrays, 0 values or nullish values:

```ts
import { isMatching, P } from 'ts-pattern';

console.log(isMatching(P.object.empty(), [])); // false
console.log(isMatching(P.object.empty(), 0)); // false
console.log(isMatching(P.object.empty(), null)); // false
console.log(isMatching(P.object.empty(), undefined)); // false
```


Lezzio:
Very nice! Looking forward to using this new pattern

yunsii:
Any plan for this PR?

nigeltroy:
I know this MR has been open for a while, but I was wondering - would users of this pattern expect `P.object` to match with anything that matches a JS `object`, or expect it to match with a variable that is a `Record` type, i.e. always a key-value map? I use the following predicate extensively for example:

```ts
/** Check if the value provided is a key-value object (not array, string, etc.) */
export const isObject = (
  value: unknown,
): value is Record<string | number | symbol, unknown> => {
  return Boolean(value?.constructor.name === 'Object');
};
```

Here, arrays and nulls return `false` while key-value objects return `true`. Personally, my mind gravitates toward using `P.array` as an Array matcher and a potential `P.object` as a key-value Object matcher. What are your thoughts?

PR #244, opened 2024-04-07
feat: added docs and test for P.object.exact(..)

JUSTIVE:


JUSTIVE:
currently, {} could be caught by `P.object.exact({a: P.any})` pattern.

JUSTIVE:
fixed {} caught by {a:P.any}, but nested object pattern should be tested

README.md
```diff
@@ -1545,6 +1545,23 @@ console.log(isMatching(P.object.empty(), null)); // false
 console.log(isMatching(P.object.empty(), undefined)); // false
 ```
 
+### `P.object.exact({...})`
+
+`P.object.exact({...})` matches objects that contain exactly the set of defined in the pattern.
```
gvergnaud, line 1550: You may have meant

```suggestion
`P.object.exact({...})` matches objects that contain exactly the set of properties defined in the pattern.
```

src/patterns.ts
```diff
@@ -656,10 +656,13 @@ function isObject(x: unknown): x is object {
...
-function hasExactKeys(keys: Set<PropertyKey>, x: unknown) {
+function hasExactKeys(keys: Set<string>, x: unknown) {  
   if (!x || typeof x !== 'object') return false;
+  const xKeys = new Set(Object.keys(x));
   if (Array.isArray(x)) return false;
+  if(xKeys.size !== keys.size) return false;
   for (const key in x) if (!keys.has(key)) return false;
+  for (const key in keys) if (!xKeys.has(key)) return false;
   return true;
 }
```
gvergnaud, line 667: 👍 makes sense

src/types/Pattern.ts
```diff
@@ -171,6 +171,10 @@ export type ObjectLiteralPattern<a> =
     }
   | never;
 
+export type ObjectExactPattern<a> = {
+  readonly [k in keyof a]: Pattern<a[k]>;
+};
```
gvergnaud, line 176: Is it to force people to include any key that exist's on an object? I think that makes a lot of sense, but how does it behave when the input has type `unknown` and `any`? Could you add tests for that?

tests/object.test.ts
```diff
@@ -106,4 +106,49 @@ describe('Object', () => {
...
+          .with(P.object.exact({ a: P.any }), (obj) => {
+            type t = Expect<Equal<typeof obj, { a: unknown }>>;
+            return 'yes';
+          })
+          .otherwise(() => 'no');
+      
+      expect(fn({a: []})).toEqual('yes');
+      expect(fn({a: null})).toEqual('yes');
+      expect(fn({a: undefined})).toEqual('yes');
+      expect(fn({a: 5})).toEqual('yes');
```
gvergnaud, line 123: Makes sure to run prettier on all these files before we merge

gvergnaud (commented):
Thanks for contributing! A few comments

tests/object.test.ts
JUSTIVE: oh, I usually apply format on save, but the formatting tool was set on biome(which is not included in this project). changed to prettier, and applied!

JUSTIVE:
during the implementation of this feature, I'm concerned about the following scenarios. 

1.  nested object, but an exact match for the outer side, and less strict matching inside.
2. recursive objects. 

maybe we need to consider those cases first before implementing it.

gvergnaud:
I was think that `exact({ ... })` would only apply to the current level of nesting. If you want everything to be exact, then you need to nest `exact`s as well:

```ts
P.object.exact({
   // this object should only have a single `a` key
   a: {
      // this object must have at least a `b` key that's a string.
      b: P.string
   }
})
```
If you want everything to be exact you'd do:

```ts
 P.object.exact({
   a: P.object.exact({
      b: P.string
   })
})

JUSTIVE:
I think that behavior could easily cause misunderstanding. I thought exact should catch 'exact' object matching, not 'exact matching, but only single depth'. It's a bit ambiguous.

JUSTIVE:
I'll keep implementing with single-depth match version(because it's a sub-spec of full-depth matching), but still not sure it's good design or not.

JUSTIVE:
@gvergnaud TIL: `typeof null is equal to 'object'`. Do you have any ideas on this case? 
image
image

Issue #144, opened 2023-02-08
Allow otherwise to chain onto exhaustive


Liam-Tait:
**Is your feature request related to a problem? Please describe.**

We have types, however they are not completely safe. I'd like to be able to ensure the exhaustive type check is run for case such as union where we want to handle where , but they also run an otherwise to prevent throwing of an error.

This means we can have the type checking where a type is partially complete.

**Describe the solution you'd like**

Add a new method `exhaustiveOtherwise` which does the exhaustive type check with a otherwise `handler`

```ts

const unit: 'days' | 'hours' = 'something'

match(unit)
.with('days', () => 'd')
.with('hours', () => 'h')
.exhaustiveOtherwise(() => 'never') // safety to prevent error
```

**Describe alternatives you've considered**

Allow `otherwise` to chain onto the end of `exhaustive`

```ts

const unit: 'days' | 'hours' = 'something'

match(unit)
.with('days', () => 'd')
.with('hours', () => 'h')
.exhaustive()
.otherwise(() => 'never')
```

Add validation such as zod to the data to handle incorrect data (hard as I would like to transition new code to be type safe without having to re-write all types)

**Additional context**

gvergnaud:
Hey! I think I'd use `.exhaustive()` and wrap the expression into a `try .. catch` to get the behavior you expect. Would that solve your problem?

eboody:
Im kind of confused about this use-case. In what context would you have a scenario that matches your example of 

`const unit: 'days' | 'hours' = 'something'`

?

Liam-Tait:
Wrapping the expressions in `try .. catch` would work, but makes the code verbose and a bit more complicated because of scoping

```ts
const value = match(unit)
  .with('days', () => 'd')
  .with('hours', () => 'h')
  .exhaustiveOtherwise(() => 'never')
```
vs
```ts
let value: string
try {
  value = match(unit)
    .with("days", () => "d")
    .with("hours", () => "h")
    .exhaustive()
} catch (error) {
  value = "never"
}
```


What I am working with is not an ideal case, the application has partially transitioned to TypeScript. Some types are only partially correct right now. 

Adding a catch or otherwise that works with exhaustive means I can start taking advantage of exhaustiveness checking, while preventing errors at runtime while types cannot be fully trusted. Maybe the ideal approach is to have everything be `unknown` and handle it that way.

[...]

PR #253, opened 2024-05-23
feat(exhaustive): Add support for passing a fallback function

gvergnaud:
Solves #144 #218 and replaces #38  #219

## Docs

By default, `.exhaustive()` will throw an error if the input value wasn't handled by any `.with(...)` clause. This should only happen if your types are incorrect.

It is possible to pass your own handler function as a parameter to decide what should happen if an unexpected value has been received. You can for example throw your own custom error:

```ts
match(...)
  .with(...)
  .exhaustive((unexpected: unknown) => {
    throw MyCustomError(unexpected);
  })
```

Or log an error and return a default value:

```ts
match<string, number>(...)
  .with(P.string, (str) => str.length)
  .exhaustive((notAString: unknown) => {
    console.log(`received an unexpected value: ${notAString}`);
    return 0;
  })
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -544,12 +544,36 @@ match(...)
   .exhaustive()
 ```
 
-Runs the pattern-matching expression and returns its result. It also enables exhaustiveness checking, making sure at compile time that we have handled all possible cases.
+Runs the pattern-matching expression and returns its result. It also enables exhaustiveness checking, making sure that we have handled all possible cases **at compile time**.
+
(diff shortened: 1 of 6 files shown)


danielsantiago:
Is there anything we can do to make this solution implemented?

Fasaloft:
Hello @gvergnaud, I have a question: What can I do to include this awesome feature in the release? It would really help me write clean and safe pattern matching even in production.

gvergnaud:
I think I'll include it in the next release. I'll try to ship it in the next few weeks

PR #257, opened 2024-06-05
feat: Added P.string.length

### LINKED ISSUE #256: P.string.length
**Is your feature request related to a problem? Please describe.**
A subpattern that can be composed with P.intersect, or P.union. .when(x=>x.length) could handle this, but it doesn't guarantee string type.
the other alternative, P.intersection(P.string.minLength(len), P.string.maxLength(len)) could be equivalent, and also can be done with APIs we already have, but it's very cumbersome and the boundaries of those min and maxes(includes or not) are confusing. 

**Describe the solution you'd like**
Add P.string.length(len: number), which matches the given value's type is string and its length is same as given `len`

**Describe alternatives you've considered**

**Additional context**

JUSTIVE:
resolves #256 .

### DIFF
--- a/src/types/Pattern.ts
+++ b/src/types/Pattern.ts
@@ -314,13 +314,13 @@ export type StringChainable<
         omitted | 'minLength'
       >;
       /**
-       * `P.string.length(len)` is a pattern, matching **strings** with at exact `len` characters.
+       * `P.string.length(len)` is a pattern, matching **strings** with exact `len` characters.
        *
        * [Read the documentation for `P.string.length` on GitHub](https://github.com/gvergnaud/ts-pattern#pstringlength)
        *
        * @example
        *  match(value)
-       *   .with(P.string.length(10), () => 'string with exact length === 10')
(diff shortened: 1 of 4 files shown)


src/types/Pattern.ts
```diff
@@ -313,6 +313,21 @@ export type StringChainable<
         MergeGuards<input, p, GuardExcludeP<unknown, string, never>>,
         omitted | 'minLength'
       >;
+      /**
+       * `P.string.length(len)` is a pattern, matching **strings** with exact `len` characters.
```
gvergnaud, line 317: did you mean
```suggestion
       * `P.string.length(len)` is a pattern, matching **strings** with exactly `len` characters.
```
?

gvergnaud (commented):
Thanks for your PR! a few minor comments

JUSTIVE:
thanks for reply! applied all of your comments

gvergnaud:
Merged and published as 5.2.0

PR #272, opened 2024-08-08
Symbols as keys

Ayc0:
I noticed that `isMatching` wasn't working with symbols:

image

But the exact same code works with strings:

image

So I updated the `matchPattern` function to also iterate through symbols in addition to keys (`Object.keys()` skips symbols, but we can use `Object.getOwnPropertySymbols()` for those)

I also added a test that is failing without my changes, but passes after:

Before | After
-|-
image | image

### DIFF
--- a/tests/objects.test.ts
+++ b/tests/objects.test.ts
@@ -1,0 +1,28 @@
+import { isMatching, P } from '../src';
+import { Equal, Expect } from '../src/types/helpers';
+
+describe('Objects', () => {
+  it('should work with symbols', () => {
+    const symbolA = Symbol('symbol-a');
+    const symbolB = Symbol('symbol-b');
+    const obj: { [symbolA]: { [symbolB]: 'foo' | 'bar' } } = {
+      [symbolA]: { [symbolB]: 'foo' },
+    };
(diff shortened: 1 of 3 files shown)


tests/objects.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    const symbolA = Symbol('symbol-a');
+    const symbolB = Symbol('symbol-b');
+    const obj: { [symbolA]: { [symbolB]: 'foo' | 'bar' } } = {
+      [symbolA]: { [symbolB]: 'foo' },
+    };
+    if (isMatching({ [symbolA]: { [symbolB]: 'foo' } }, obj)) {
+      type t = Expect<Equal<typeof obj, { [symbolA]: { [symbolB]: 'foo' } }>>;
+    } else {
+      throw new Error('Expected obj to match the foo pattern!');
+    }
```
Ayc0, line 15: This test only wasn't enough as it was passing before (as it was stripping all symbols)

So I added the other `if` for the negative check

gvergnaud:
Awesome, thank you!

gvergnaud (approved):

Issue #271, opened 2024-08-08
Match + exhaustive doesn't work with readonly array


rikutiira:
**Describe the bug**
Hey, first of all thanks for the awesome library 🙏  I have a small issue where `.match()` with `.exhaustive()` is not working with a readonly array, I included screenshots of the issue.

```ts
match(value) // value: string | Date | readonly string[]
  .with(P.string, (value) => ...)
  .with(P.array(P.string), (value) => ...)
  .with(P.instanceOf(Date), (value) => ...)
  .exhaustive() // Complains about "Type 'NonExhaustiveError<readonly string[]>' has no call signatures."
```

Interestingly enough, the value is correct inside the `P.array(P.string))` condition: `value: string[] | readonly string[]`.

**TypeScript playground with a minimal reproduction case**
https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbziAhjAxgCwDRwApwC+cAZlBCHAOQwDOAtGGjAKZQB2VAUFzAJ5gWcAGooANgFchAXji0YUYOwDmcAD5woLFABMI7MXzkKlygNoBddXAAiaFgG4eOlujEotcdPvlwAbuJSAFwigY48qBiYABQBkiwAlFxwKXAAdADuwDAxeGnyiiq4sWEJcNIAfHAAjEmp6Vk50XkeUCh8zfkmKgnFcVJllXAATHWpmdm5aUryKOzoLADyJNF2rL1wJfGDVQDMYylpLAAemCgS8sB+LNFJQA

**Versions**
- TypeScript version: 5.4.2
- ts-pattern version: 5.2.0
- environment: node v18.15.0

<img width="656" alt="exhaustive_readonly_array_1" src="https://github.com/user-attachments/assets/2f2c84ac-5777-47de-a1a4-f76c7316e0d4">
<img width="955" alt="exhaustive_readonly_array_3" src="https://github.com/user-attachments/assets/3b2f5fc8-2866-4531-b30f-f8af25e635cd">
<img width="1132" alt="exhaustive_readonly_array_2" src="https://github.com/user-attachments/assets/5a1a4704-6ffe-471d-a078-4b10be0adf1b">

PR #284, opened 2024-09-20
Fixes type `InvertPatternForExcludeInternal` to work with readonly array

changwoolab:
This PR resolves the issue https://github.com/gvergnaud/ts-pattern/issues/271

Fixed the issue by modifying `InvertPatternForExcludeInternal` to consider `readonly` .

### DIFF
--- a/src/types/InvertPattern.ts
+++ b/src/types/InvertPattern.ts
@@ -318,9 +318,10 @@ type InvertPatternForExcludeInternal<p, i, empty = never> =
     ? {
         select: InvertPatternForExcludeInternal<subpattern, i, empty>;
         array: i extends readonly (infer ii)[]
-          ? i extends any[]
-            ? InvertPatternForExcludeInternal<subpattern, ii, empty>[]
-            : readonly InvertPatternForExcludeInternal<subpattern, ii, empty>[]
+          ? MaybeAddReadonly<
+              InvertPatternForExcludeInternal<subpattern, ii, empty>[],
+              IsReadonlyArray<i>
+            >
           : empty;
         map: subpattern extends [infer pk, infer pv]
           ? i extends Map<infer ik, infer iv>
--- a/src/types/InvertPattern.ts
+++ b/src/types/InvertPattern.ts
@@ -318,7 +318,9 @@ type InvertPatternForExcludeInternal<p, i, empty = never> =
     ? {
         select: InvertPatternForExcludeInternal<subpattern, i, empty>;
         array: i extends readonly (infer ii)[]
-          ? InvertPatternForExcludeInternal<subpattern, ii, empty>[]
+          ? i extends any[]
+            ? InvertPatternForExcludeInternal<subpattern, ii, empty>[]
+            : readonly InvertPatternForExcludeInternal<subpattern, ii, empty>[]
           : empty;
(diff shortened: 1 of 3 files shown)


src/types/InvertPattern.ts
```diff
@@ -318,7 +318,9 @@ type InvertPatternForExcludeInternal<p, i, empty = never> =
     ? {
         select: InvertPatternForExcludeInternal<subpattern, i, empty>;
         array: i extends readonly (infer ii)[]
-          ? InvertPatternForExcludeInternal<subpattern, ii, empty>[]
+          ? i extends any[]
+            ? InvertPatternForExcludeInternal<subpattern, ii, empty>[]
+            : readonly InvertPatternForExcludeInternal<subpattern, ii, empty>[]
```
gvergnaud, line 323: **Stylistic**: There's a pattern of using `MaybeAddReadonly<T>` in this file that I think we should keep:

```suggestion
          ? MaybeAddReadonly<
              InvertPatternForExcludeInternal<subpattern, ii, empty>[],
              IsReadonlyArray<i>
            >
```

tests/lists.test.ts
```diff
@@ -59,6 +59,17 @@ describe('List ([a])', () => {
...
+  it('issue #271: P.array should support readonly arrays as its input', () => {
+    type Input = string | Date | readonly string[];
+    const input = ['a', 'b', 'c'] as Input;
+
+    const output = match(input)
+      .with(P.array(P.string), (value) => 2)
+      .with(P.string, (value) => 1)
+      .with(P.instanceOf(Date), (value) => 3)
+      .exhaustive();
+  });
```
gvergnaud, line 71: Could you move this unit test to the `exhaustive-match.test.ts` file?

I think this is technically a bug of exhaustiveness checking, since the problem is that without this change, the match expression isn't considered exhaustive.

gvergnaud (commented):
Great investigation! a few minor comments but otherwise it looks good to me

tests/lists.test.ts
changwoolab: @gvergnaud Sure! I moved it in this commit [980e4d0](https://github.com/gvergnaud/ts-pattern/pull/284/commits/980e4d085e0e55bb7e3d44f715f5fa9f63e5e66b)

src/types/InvertPattern.ts
changwoolab: @gvergnaud 
it looks good! Included it in this commit [979b358](https://github.com/gvergnaud/ts-pattern/pull/284/commits/979b358fbb5fbe14cc2af9d6da1d09bcfad18830)

tests/exhaustive-match.test.ts
```diff
@@ -979,4 +979,15 @@ describe('exhaustive()', () => {
       expect(withoutTypo({ sex: 'b', age: 'c' })).toBe(1);
     });
   });
+
+  it('issue #271: P.array should support readonly arrays as its input', () => {
```
gvergnaud, line 983: Nit: I think this would be clearer:
```suggestion
  it('issue #271: P.array should exhaustively match readonly arrays', () => {
```

gvergnaud (approved):
Thanks!

PR #287, opened 2024-09-24
fix matchPatter helper code structures

kirylbehan:
I always check source code of the package before using it.
Some problematic structures rushed in, here's a fix if you don't mind .
Thanks

### DIFF
--- a/src/internals/helpers.ts
+++ b/src/internals/helpers.ts
@@ -53,8 +53,7 @@ export const matchPattern = (
       let endPatterns = [];
       let variadicPatterns: AnyMatcher[] = [];
 
-      for (const i of pattern.keys()) {
-        const subpattern = pattern[i];
+      for (const subpattern of pattern) {
         if (isMatcher(subpattern) && subpattern[symbols.isVariadic]) {
           variadicPatterns.push(subpattern);
         } else if (variadicPatterns.length) {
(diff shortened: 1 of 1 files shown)


src/internals/helpers.ts
```diff
@@ -90,9 +89,7 @@ export const matchPattern = (
           endPatterns.every((subPattern, i) =>
             matchPattern(subPattern, endValues[i], select)
           ) &&
-          (variadicPatterns.length === 0
-            ? true
-            : matchPattern(variadicPatterns[0], middleValues, select))
+          matchPattern(variadicPatterns[0], middleValues, select)
```
gvergnaud, line 92: This is incorrect and would likely break test on tuple pattern.

gvergnaud (commented):
Hey, I don't accept stylistic code contributions because everyone has different preferences and there is no way any code can esthetically please everyone. I don't think it's worth it. 

If there are specific behaviors you think are bugs and would like to fix, please open an issue first

src/internals/helpers.ts
kirylbehan: above check `if (variadicPatterns.length) {`

kirylbehan:
Its less code, not esthetically better but clear and modern
As you please

Issue #291, opened 2024-10-02
TSError: The inferred type of 'MY_PATTERN' cannot be named without a reference to '<path>/ts-pattern/dist/types/Pattern'


GGomez99:
**Describe the bug**

When working in a yarn project with many workspaces, and having this `compilerOptions` in their `tsconfig.json`:
```json
"compilerOptions": {
   "moduleResolution": "bundler"
}
```

I get a TS error when I try to directly export a `Pattern` without explicitly typing its type:
```
The inferred type of 'MY_PATTERN' cannot be named without a reference to '../.yarn/cache/ts-pattern-npm-5.4.0-2464380e11-f757dd3259.zip/node_modules/ts-pattern/dist/types/Pattern'. This is likely not portable. A type annotation is necessary.
```

**TypeScript playground with a minimal reproduction case**

Example: [Playground](https://codesandbox.io/p/devbox/r2jp93?file=%2Fmy-package%2Fexample.ts%3A9%2C21)

**Versions**
- TypeScript version: 5.6.2
- ts-pattern version: 5.4.0
- environment: browser / node version

**Possible solutions**

In order to solve this, we could:
1. Directly export the needed types in the `package.json` of `ts-pattern`
2. Put the needed types in a namespace, use it in all the other files and then export that namespace

I think the first solution is the simplest and it won't modify the inferred type names when hovering any ts-pattern functions. The only drawback is that internal types will be exposed through `ts-pattern/types` but I think that's worth it. I'll make a PR for this!

lucaschultz:
Having the same problem; I'm missing the `Chainable` and `GuardP` types to be able to create a reusable pattern. @gvergnaud, should I whip up another PR, or is someone already on this?

lucaschultz:
Ah, I think I just realized that there hasn't been a release since this PR was merged. That's probably why I'm having the problem.

gvergnaud:
@GGomez99 's PR has been published as v5.5.0! Closing this issue

PR #292, opened 2024-10-03
Export Pattern types in package.json

GGomez99:
Fix issue: https://github.com/gvergnaud/ts-pattern/issues/291

## Changes

Export `Pattern.d.ts` in the package.json to solve the issue. They are now available at `ts-pattern/types`.

## QA

Doing the same changes from this PR in [this playground](https://codesandbox.io/p/devbox/r2jp93?file=%2Fmy-package%2Fexample.ts%3A9%2C21) fixes the TS error

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,14 @@
       "types": "./dist/index.d.ts",
       "default": "./dist/index.js"
     },
+    "./types": {
+      "require": {
+        "types": "./dist/types/Pattern.d.cts"
+      },
+      "import": {
+        "types": "./dist/types/Pattern.d.ts"
+      }
+    },
     "./package.json": "./package.json"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,6 +17,14 @@
...
       "default": "./dist/index.js"
     },
+    "./types": {
+      "require": {
+        "types": "./dist/types/Pattern.d.cts"
+      },
+      "import": {
+        "types": "./dist/types/Pattern.d.ts"
+      }
+    },
```
gvergnaud, line 27: Could you create a `types/index.ts` file that exports the types we need from `Pattern.ts`? This should make it easier to export type from other files if necessary in the future

PR #293, opened 2024-10-04
Export additional useful types

trevorsmith:
I'd like access to more internal ts-pattern types to:

- Create wrapper functions around `match` (e.g. from a class method)
- Type generics passed into `match` / `when` / etc.
- Explicitly type matches inside other types, like interfaces
- etc.

I've included exports all of the potential user-facing types I could find – did I miss any?


svemat01:
What's the reason for including the dist folder in the repo?

PR #297, opened 2024-11-01
ci(bench): migrate benchmark tool benny to vitest

sunrabbit123:
# The reason why I raised the PR is as follows.

1. [Benchmark order affects results](https://github.com/caderek/benny/issues/57)
2. [if function result is not used, this function is not using](https://stackoverflow.com/questions/65045683/invocation-of-javascript-class-methods-is-20-40-times-slower-than-same-function#:~:text=%27empty%2Dbody%2Dtest%27%20is%20the%20same%20the%20as%20with%20invocation%20of%20dummy%20method)
3. Disclosure of incorrect information due to incorrect benchmarks

## Proposal

1. remove benny
2. and use vitest

스크린샷 2024-11-01 오후 3 52 55

If this PR is accepted, there is also a foothold to replace all tests with bitest-based rather than just test.


gvergnaud:
Interesting, so because of the ordering of test cases, the current benchmarks where biased toward native implementation and vitest's results show that ts-pattern's performance are more similar to native control flow statements. Am I interpreting this correctly?

benchmarks/nested-objects.bench.ts
```diff
@@ -0,0 +1,74 @@
...
+      0: { type: 'a' as const, value: { x: Math.random(), y: Math.random() } },
+      1: {
+        type: 'b' as const,
+        value: Math.random() > 0.5 ? [1, 2, 3, 4] : ['hello'],
+      },
+      2: { type: 'c' as const, age: Math.random(), name: 'acdfl' },
+    };
+  
+    return map[inputIndex];
+  })();
```
gvergnaud, line 18: We should re-generate the input on every run instead of only once, otherwise we are always testing against the same object which could make results vary between runs, and could result in irrelevant JIT optimizations.

benchmarks/random-digit.bench.ts
```diff
@@ -0,0 +1,131 @@
+import { bench, describe, expect } from 'vitest';
+import { match } from '../src';
+
+type Digit = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9;
+
+const rand = () => Math.floor(Math.random() * 10) as Digit;
+
+describe('ts-pattern-benchmark/random-digit', () => {
+  const digit = rand();
```
gvergnaud, line 9: same here

vitest.config.mts
```diff
@@ -0,0 +1,12 @@
...
+
+export default defineConfig({
+  test: {
+    name,
+    coverage: {
+      provider: 'istanbul',
+      include: ['src/**/*'],
+    },
+  },
+});
```
gvergnaud, line 12: Could we move this to the benchmark folder to make sure it's scope to it?

sunrabbit123: I put it as the default setting for the bitest, but if you don't want to see it, you don't need it

benchmarks/nested-objects.bench.ts
sunrabbit123: Nevertheless, the object creation time should be excluded from the benchmark.

Therefore, I am looking for ways to use setup.

I will ask you to review it again after working on it.

sunrabbit123:
> interesting, so because of the ordering of test cases, the current benchmarks where biased toward native implementation and vitest's results show that ts-pattern's performance are more similar to native control flow statements. Am I interpreting this correctly?

What is certain is that Benny is a problematic benchmark tool.

As for whether the beatest is correct, I think it is better to check the benchmark result one more time after modifying it according to the review.

benchmarks/nested-objects.bench.ts
gvergnaud: I don't mind that the object creation is included in the benchmark as long as it's included in all cases. Its only purpose is to compare equivalent implementation of the same thing, so the absolute values aren't that important

sunrabbit123: Even if it's an absolute value, it seems to affect the benchmark results.
It doesn't affect who's faster or who's slower
There's definitely an impact when it comes to showing how much faster it is, how much slower it is.

Of course, even if the impact is minimal due to the complexity of the benchmark content.

Is there any problem with JIT about randomly importing pre-generated objects for each benchmark?
i want to import pre-generated objects using random values.

sunrabbit123:
next is changed benchmark result

스크린샷 2024-11-04 오전 11 45 56

I still don't feel the big gap.
What's the problem..

PR #325, opened 2025-06-25
fix(match): otherwise should now support correctly discriminated union

Thenkei:


gvergnaud:
Closing in favor of using `.narrow()` before your otherwise. see https://github.com/gvergnaud/ts-pattern/releases/tag/v5.8.0

